PR #5, opened 2023-07-05
feat(twap): ✨ add initial version of twap service

tukantje:
# How to test?
* yarn
* yarn compose:up

This should allow you to have everything you need up and running. Please note you need to have docker present in your machine.

Please do ignore vercel issues, as we should be getting rid of that after this is deployed to production.


apps/twap/README.md
```diff
@@ -0,0 +1,7 @@
+# Awesome Project Build with TypeORM
+
+Steps to run this project:
+
+1. Run `npm i` command
+2. Setup database settings inside `data-source.ts` file
+3. Run `npm start` command
```
alfetopito, line 7: `yarn`?

apps/twap/project.json
```diff
@@ -0,0 +1,116 @@
...
+      "executor": "nx:run-commands",
+      "outputs": [
+        "{options.outputPath}"
+      ],
+      "options": {
+        "command": "ts-node --project tsconfig.app.json ../../node_modules/.bin/typeorm migration:generate --pretty -d datasource.config.ts src/migrations/migration",
+        "cwd": "apps/twap"
+      }
+    },
+    "typeorm-run-migrations": {
```
alfetopito, line 104: Nitpick: Could we maybe omit the tool name from the commands? (remove `typeorm-` prefix)

apps/twap/project.json
```diff
@@ -0,0 +1,116 @@
...
+      "executor": "nx:run-commands",
+      "outputs": [
+        "{options.outputPath}"
+      ],
+      "options": {
+        "command": "ts-node --project tsconfig.app.json ../../node_modules/.bin/typeorm migration:run -d datasource.config.ts",
```
alfetopito, line 110: Nitpick: feels very verbose.
Could we create an alias for running typeomr?

apps/twap/src/app/app.spec.ts
```diff
@@ -0,0 +1,20 @@
...
+    server.register(app);
+  });
+
+  it('should respond with a message', async () => {
+    const response = await server.inject({
+      method: 'GET',
+      url: '/',
+    });
+
+    expect(response.json()).toEqual({ message: 'Hello API' });
```
alfetopito, line 18: This test fails, right?

apps/twap/src/app/data/order.ts
```diff
@@ -0,0 +1,87 @@
...
+      buyToken: this.buyToken,
+      receiver: this.receiver,
+      partSellAmount: this.partSellAmount,
+      minPartLimit: this.minPartLimit,
+      t0: this.t0,
+      n: this.n,
+      t: this.t,
+      span: this.span,
+      appData: this.appData,
+    });
```
alfetopito, line 32:
```suggestion
```

apps/twap/src/app/data/order.ts
```diff
@@ -0,0 +1,87 @@
...
+  buyToken: string;
+
+  @Column('varchar')
+  appData: string;
+
+  @Column('varchar')
+  receiver: string;
+
+  @Column('varchar')
+  chainId: number;
```
alfetopito, line 68: Wait, you map here as a number but still use a `varchar`?

apps/twap/src/app/plugins/env.ts
```diff
@@ -0,0 +1,53 @@
...
+    DATABASE_USERNAME: {
+      type: 'string',
+    },
+    DATABASE_PASSWORD: {
+      type: 'string',
+    },
+  },
+};
+
+export default fp(async (fastify, opts) => {
```
alfetopito, line 32: What does `fp` stands for?
A special name in the freamework?

apps/twap/src/app/plugins/infura.ts
alfetopito, line 1: This is still being worked on as you mentioned, right?

apps/twap/src/app/routes/_chainId/_walletAddress/orders/index.ts
```diff
@@ -0,0 +1,185 @@
+import { FastifyInstance } from 'fastify';
+import { Wallet } from '../../../../data/wallet';
+import { Order } from '../../../../data/order';
```
alfetopito, line 3: Nipick: does it make sense to do full path imports to make it look cleaner?

apps/twap/src/app/routes/_chainId/_walletAddress/orders/index.ts
```diff
@@ -0,0 +1,185 @@
...
+import { FromSchema, JSONSchema } from 'json-schema-to-ts';
+
+const ADDRESS_LENGTH = 42;
+
+const routeSchema = {
+  type: 'object',
+  required: ['chainId', 'walletAddress'],
+  properties: {
+    chainId: {
+      enum: ['1', '5', '100'],
```
alfetopito, line 13: Are the networks used anywhere else?
Would it make sense to come from a central config location, or even env vars?

apps/twap/src/app/routes/_chainId/_walletAddress/orders/index.ts
```diff
@@ -0,0 +1,185 @@
...
+          sellToken,
+          buyToken,
+          receiver,
+          partSellAmount,
+          minPartLimit,
+          t0,
+          n,
+          t,
+          span,
+          appData,
```
alfetopito, line 143: Do you have to deconstruct all fields?
Would it be too bad to simply spread them onto the `create` method?

apps/twap/src/app/routes/_chainId/_walletAddress/orders/index.ts
```diff
@@ -0,0 +1,185 @@
...
+        },
+        safeTxHash,
```
alfetopito, line 145: `safeTxHash` doesn't seem to be used anywhere

apps/twap/src/app/routes/root.ts
```diff
@@ -0,0 +1,10 @@
+import { FastifyInstance, FastifyRequest, FastifyReply } from 'fastify';
+
+export default async function (fastify: FastifyInstance) {
+  fastify.get(
+    '/',
+    async function (request: FastifyRequest, reply: FastifyReply) {
+      return { message: 'Hello API' };
+    }
```
alfetopito, line 8: Aha!
The test won't fail then as we have this here.

Remove?

libs/abis/README.md
```diff
@@ -0,0 +1,23 @@
+# CoW Swap Abis
+
+Exports relevant ABIs )types and JSON)
```
alfetopito, line 3: Typo

```suggestion
Exports relevant ABIs (types and JSON)
```

libs/abis/project.json
```diff
@@ -0,0 +1,53 @@
...
+      "dependsOn": [
+        "abi:custom"
+      ]
+    },
+    "abi:ethflow": {
+      "executor": "nx:run-commands",
+      "options": {
+        "command": "typechain --target ethers-v5 --out-dir ./libs/abis/src/generated/ethflow \"./node_modules/@cowprotocol/ethflowcontract/artifacts/CoWSwapEthFlow.sol/*.json\""
+      }
+    },
```
alfetopito, line 18: You can remove this

```suggestion
```

libs/abis/project.json
```diff
@@ -0,0 +1,53 @@
...
+      "options": {
+        "outputPath": "dist/libs/abis"
+      }
+    },
+    "publish": {
+      "command": "node tools/scripts/publish.mjs abis {args.ver} {args.tag}",
+      "dependsOn": [
+        "build"
+      ]
+    },
```
alfetopito, line 39: Probably remove this one too

```suggestion
```

package.json
```diff
@@ -3,46 +3,69 @@
...
+    "serve": "nx serve",
+    "build": "nx build",
+    "create-app": "nx g app --framework=fastify --docker",
     "docker-build:affected": "nx affected -t docker-build --prod --parallel --maxParallel=3 --skip-nx-cache --skip-nx-cache",
     "compose:up": "npm run docker-build:affected && docker-compose up"
   },
   "private": true,
   "dependencies": {
-    "@fastify/autoload": "^5.0.0",
+    "@cowprotocol/ethflowcontract": "cowprotocol/ethflowcontract.git#v1.0.0-artifacts",
```
alfetopito, line 16: Probably don't need this since you chopped the abis

alfetopito (commented):
Left a few comments 😬

apps/twap/README.md
tukantje: Autogenerated filed. Replaced npm with yarn in general.

apps/twap/src/app/data/order.ts
tukantje: Good one - changed it.

apps/twap/src/app/plugins/env.ts
tukantje: fastify plugin. it is a default export you can name it as you wish, but conventionally this is the naming used.

apps/twap/src/app/routes/_chainId/_walletAddress/orders/index.ts
tukantje: Can be a later follow up. Don't have capacity for this at the moment.

apps/twap/src/app/plugins/infura.ts
tukantje: Yes - safe to ignore for this moment.

apps/twap/src/app/routes/_chainId/_walletAddress/orders/index.ts
tukantje: Updated it.

Personally, it is more explicit - while we can write this with a spread, that tends to hide information which is why I don't prefer it.

apps/twap/src/app/routes/_chainId/_walletAddress/orders/index.ts
tukantje: Removed. Will be used later.

apps/twap/src/app/routes/root.ts
tukantje: Updated to a health check endpoint. It is good practice to have an endpoint that returns just a flat response, to be able to easily test if we are live with the service or not (also later on for automated tests).

apps/twap/src/app/app.spec.ts
tukantje: Replaced with health check's test.

apps/twap/src/app/routes/_chainId/_walletAddress/orders/index.ts
tukantje: In principle I agree, but would prefer to circle back on this to avoid premature abstraction. There are a couple of questions to be answered here since we cannot treat all the chains the same.

Env vars won't work in general though, as we are even defining the schema as _const_ to be inlined in build step. So probably still somewhere in code.

apps/twap-e2e/src/twap/twap.spec.ts
```diff
@@ -0,0 +1,10 @@
+import axios from 'axios';
+
+describe('GET /', () => {
+  it('should return a message', async () => {
```
alfetopito, line 4: Nitpick:

```suggestion
  it('should be healthy', async () => {
```

apps/twap/src/migrations/1688633087604-migration.ts
```diff
@@ -0,0 +1,31 @@
...
+export class Migration1688633087604 implements MigrationInterface {
+  name = 'Migration1688633087604';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`
+            ALTER TABLE "order" DROP COLUMN "chainId"
+        `);
+    await queryRunner.query(`
+        TRUNCATE TABLE "order"
+        `);
```
alfetopito, line 12: Does it need to truncate the whole table?

Anyway, doesn't actually matter as this is not production ready yet.

Disregard

alfetopito: Nitpick:

```suggestion
    await queryRunner.query(`
            TRUNCATE TABLE "order"
        `);
```

libs/abis/src/generated/ethflow/CoWSwapEthFlow.ts
alfetopito, line 1: Just noticed, this file probably shouldn't be here, since we don't have ethflow stuff.
Same for the whole folder.

apps/twap/src/app/plugins/env.ts
alfetopito: No need, just wanted to understand where it came from.

apps/twap/src/app/routes/_chainId/_walletAddress/orders/index.ts
alfetopito: My personal preference is to pass along without deconstructing because:
1. One more place to (forget to) update when adding a new field to the API
2. If the API enforces no fields not defined in the schema, there shouldn't be anything weird coming in.
3. The schema is the single source of truth. Still additional transformations can occur here, but the bulk of the fields remain what the API expects

alfetopito (approved):

apps/twap/src/app/plugins/infura.ts
```diff
@@ -0,0 +1,41 @@
+import { SupportedChainId } from '@cowprotocol/cow-sdk';
+import { ethers } from 'ethers';
+import { FastifyInstance } from 'fastify';
+import { ComposableCoW__factory } from '@cow-web-services/abis';
+import fp from 'fastify-plugin';
+
+const NETWORKS = ['goerli'];
```
nenadV91, line 7: Is `goerli` only here intended?
What about other networks?

apps/twap/src/app/plugins/infura.ts
```diff
@@ -0,0 +1,41 @@
...
+export const COMPOSABLE_COW_ADDRESS: Record<SupportedChainId, string> = {
+  1: '0xF487887DA5a4b4e3eC114FDAd97dc0F785d72738',
+  100: '0xF487887DA5a4b4e3eC114FDAd97dc0F785d72738',
+  5: '0xF487887DA5a4b4e3eC114FDAd97dc0F785d72738',
+};
+
+const SINCE = 9289470;
```
nenadV91, line 14: Why this block specifically?
If there is a reason, we could add some comment as description.

apps/twap/src/app/utils/getConditionalOrderId.ts
```diff
@@ -0,0 +1,85 @@
...
+  // timeInterval
+  t: number;
+  span: number;
+  appData: string;
+}
+
+export const TWAP_ORDER_STRUCT =
+  'tuple(address sellToken,address buyToken,address receiver,uint256 partSellAmount,uint256 minPartLimit,uint256 t0,uint256 n,uint256 t,uint256 span,bytes32 appData)';
+
+const twapHandlerAddress = '0x910d00a310f7Dc5B29FE73458F47f519be547D3d';
```
nenadV91, line 43: What is this exactly?

apps/twap/src/migrations/1688472763977-migration.ts
```diff
@@ -0,0 +1,47 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class Migration1688472763977 implements MigrationInterface {
+    name = 'Migration1688472763977'
```
nenadV91, line 4: Is this being auto-generated?
Would it be possible to add some more descriptive names or at least some comments that describe what each of these migrations does?

tukantje: They are all autogenerated, but you can edit them. In principle, you can add more info as they are generated, but I would suggest relying on git history on that regard.

Names must always follow a "${name}${timestamp}" fashion though so would be like "SuperInformativeName1231283781".

PR #7, opened 2023-07-14
feat(orderbook): ✨ add orderbook integration

tukantje:
This PR adds orderbook integration, and data models & plugins for it.

As per `plugins/orderbook.ts` shows, we create an orderbook plugin, which is decorating the fastify instance as per line 85. This means that you can, once the plugin is instantiated, in any part of the application use it as;
```
fastify.orderbook.goerli.settlement (followed by ORM calls like find, findOne etc.)
fastify.orderbook.goerli.trade (followed by ORM calls like find, findOne etc.)
fastify.orderbook.goerli.order (followed by ORM calls like find, findOne etc.)
fastify.orderbook.goerli.getExecutionInfo - currently not a db call per se, but is intended to be. Will follow up on this one to see if I can read through production DB, but can't do that in dev mode, on my device.
```

Under `utils/transformers.ts` you will notice some helpers, these are used to marshall / unmarshall data from / to database since orderbook database makes heavy use of BYTEA data type.

For relevant .env file definitions, please refer to message on main slack channel or feel free to reach out to me.


apps/twap/src/app/orderbook/order.ts
```diff
@@ -0,0 +1,126 @@
...
+  sellAmount: number;
+
+  @Column('numeric', { name: 'buy_amount' })
+  buyAmount: number;
+
+  @Column('bigint', {
+    name: 'valid_to',
+    transformer: { from: stringToBigInt, to: bigIntToString },
+  })
+  validTo: bigint;
```
alfetopito, line 48: Is there any format we can use for timestamp?

apps/twap/src/app/orderbook/order.ts
```diff
@@ -0,0 +1,126 @@
...
+    name: 'buy_token',
+    transformer: { from: bufferToString, to: stringToBuffer },
+  })
+  buyToken: string;
+
+  @Column('numeric', { name: 'sell_amount' })
+  sellAmount: number;
+
+  @Column('numeric', { name: 'buy_amount' })
+  buyAmount: number;
```
alfetopito, line 42: Shouldn't the amounts be bigints as well?

apps/twap/src/app/orderbook/trade.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  @Column('bytea', {
+    name: 'order_uid',
+    transformer: { from: bufferToString, to: stringToBuffer },
+  })
+  orderUid: string;
+
+  @Column('numeric', { name: 'sell_amount' })
+  sellAmount: string;
+}
```
alfetopito, line 31: Is there no buy amount as part of the trade?

apps/twap/src/app/plugins/orderbook.ts
```diff
@@ -0,0 +1,110 @@
...
+    // Based on: https://github.com/cowprotocol/services/blob/main/crates/database/src/orders.rs#L483-L549
+    // Other options are: Subgraph or watch for the settlement event
+    try {
+      const orders: ExecutionInfo[] = await Promise.all(
+        partIds.map((partId) =>
+          fetch(`${apiBaseUrl}/orders/${partId}`)
+            .then((res) => res.json())
+            .then((json) => ({
+              executedBuyAmount: BigInt(json.executedBuyAmount),
+              executedSellAmount: BigInt(json.executedSellAmount),
```
alfetopito, line 35: You might want to use `executedSellAmountBeforeFees` instead.

Maybe it doesn't matter for non-swap orders, where the fee is taken from the surplus, but still, keep that in mind.

See this TWAP part order example https://api.cow.fi/goerli/api/v1/orders/0x5173a80aa1d97bceb2e0a72ea7ac1de47825c3b157c4c443ae2ec6415607ba385895fcf0a96a4d19e17d2dd759926d1a02781ef864b14a83

Both are the same, but depending on what you querying it won't be.

apps/twap/src/app/plugins/orderbook.ts
```diff
@@ -0,0 +1,110 @@
...
+              executedFeeAmount: BigInt(json.executedFeeAmount),
```
alfetopito, line 36: Are you sure this is what you are looking for?

For non-swap orders, this will always be 0 AFAICT.

In the SDK we have one transformation step to fill this the way one would expect 

https://github.com/cowprotocol/cow-sdk/blob/2e5cc9df8cad563ae3f51a2aedf6efe22dbc16e6/src/order-book/transformOrder.ts#L16

See the example TWAP part order in the previous comment
Screenshot 2023-07-14 at 16 26 08

apps/twap/src/app/plugins/orderbook.ts
```diff
@@ -0,0 +1,110 @@
...
+      getExecutionInfo: getExecutionInfoFactory(
+        fastify.orm['mainnet'],
+        getApiBaseUrl(SupportedChainId.MAINNET)
+      ),
+    };
+    fastify.decorate('orderbook', {
+      goerli,
+      mainnet,
+      [SupportedChainId.GOERLI]: goerli,
+      [SupportedChainId.MAINNET]: mainnet,
```
alfetopito, line 89: Nitpick: this looks repetitive and for now there are only 2 networks, but in the future it might not.
Would it be possible to pass only a list of networks and build it all from there?

apps/twap/src/app/orderbook/order.ts
tukantje: image

DB definition on this is numeric, so not particularly certain. As opposed to valid_to, which is bigint.

apps/twap/src/app/orderbook/order.ts
tukantje: image
Not defined as a timestamp, as opposed to `creation_timestamp`. We can later on add transformers to these, but as of right now it might be unnecessary.

apps/twap/src/app/orderbook/trade.ts
tukantje: image
I must have missed it, fixing it.

apps/twap/src/app/plugins/orderbook.ts
tukantje: Good point, updating this. cc/ @shoom3301 for visibility

apps/twap/src/app/plugins/orderbook.ts
tukantje: It is a good point, I've updated this to have a factory pattern.

apps/twap/src/app/plugins/orderbook.ts
shoom3301: Leandro is right. But, for TWAP frontend we don't need this value. For now we can just skip it at all

apps/twap/src/app/plugins/orderbook.ts
```diff
@@ -0,0 +1,112 @@
...
+              accumulator.executedSellAmount + BigInt(order.executedSellAmount),
+            executedFeeAmount:
+              accumulator.executedFeeAmount + BigInt(order.executedFeeAmount),
+          };
+        },
+        {
+          executedBuyAmount: BigInt(0),
+          executedSellAmount: BigInt(0),
+          executedFeeAmount: BigInt(0),
+        }
```
shoom3301, line 57: { ...DEFAULT_TWAP_EXECUTION_INFO }

shoom3301, line 57:
```suggestion
        { ...DEFAULT_TWAP_EXECUTION_INFO }
```

apps/twap/src/app/plugins/orderbook.ts
shoom3301: However, in general we should use the transformation here

apps/twap/src/app/orderbook/order.ts
```diff
@@ -0,0 +1,126 @@
...
+  })
+  buyTokenBalance: 'erc20' | 'internal';
+
+  @Column('numeric', { name: 'full_fee_amount' })
+  fullFeeAmount: number;
+
+  @Column('enum', {
+    name: 'class',
+    enumName: 'orderclass',
+    enum: ['market', 'liquidity', 'limit'],
```
nenadV91, line 117: Do you think we could use `import { OrderClass } from '@cowprotocol/cow-sdk'` here?
It would be easier to change it, if its in one place, but I will leave it up to you.

apps/twap/src/app/orderbook/settlement.ts
```diff
@@ -0,0 +1,45 @@
...
+  })
+  blockNumber: bigint;
+
+  @PrimaryColumn('bigint', {
+    name: 'log_index',
+    transformer: { from: stringToBigInt, to: bigIntToString },
+  })
+  logIndex: bigint;
+
+  @Column('bytea', {
```
nenadV91, line 23: Should we add a `name` prop here also?

nenadV91: Also what is this field actually?
Name of the solver?

nenadV91 (approved):
Left some minor comments.
Nice work 👍

apps/twap/src/app/plugins/orderbook.ts
tukantje: Will update this - thank you :+1:

apps/twap/src/app/orderbook/order.ts
```diff
@@ -0,0 +1,126 @@
...
+  bufferToString,
+  stringToBigInt,
+  stringToBuffer,
+} from '../utils/transformers';
+
+@Entity({ name: 'orders' })
+export class Order {
+  @PrimaryColumn('bytea', {
+    name: 'uid',
+    transformer: { from: bufferToString, to: stringToBuffer },
```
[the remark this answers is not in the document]
tukantje: Other way around.

Database types for this table is not controlled by us. It is defined as `bytea` which is a byte array, in JS that would be a buffer.

We process this buffer into a string to be able to use it on our side. It gets converted back into buffer before being written to db.

apps/twap/src/app/orderbook/order.ts
tukantje: It would be the wrong abstraction, because this definition has to follow what DB has defined (which is a read only replica of orderbook DB). Since we don't control their structure, we can't be certain that this will always match on cow-sdk as well.

apps/twap/src/app/orderbook/settlement.ts
tukantje: Address of the solver, from what I can see in the DB.

We can't really adjust the data type for these, since they are what backend team has in place. Entities under orderbook table exist to give us a way to read the DB itself, essentially, so it just defines what is in place.

apps/twap/src/app/plugins/orderbook.ts
tukantje: Good one - added it

apps/twap/src/app/utils/getApiBaseUrl.ts
```diff
@@ -0,0 +1,15 @@
...
+const COW_API_BASE_URL = 'https://api.cow.fi';
+
+const CHAIN_ID_TO_NAME: Record<SupportedChainId, string> = {
+  1: 'mainnet',
+  5: 'goerli',
+  100: 'xdai',
+};
+
+export function getApiBaseUrl(chainId: SupportedChainId): string {
+  const chainName = CHAIN_ID_TO_NAME[chainId];
```
anxolin, line 12: why not using the SDK? so you get nice types and we dog-food our own products?

.env.example
```diff
@@ -5,4 +5,8 @@ INFURA_KEY=infura-key
 INFURA_JWT_ID=infura-jwt-id
 DATABASE_USERNAME=bff-db-user
 DATABASE_PASSWORD=bff-db-password
-DATABASE_NAME=bff-db
\ No newline at end of file
+DATABASE_NAME=bff-db
+ORDERBOOK_DATABASE_HOST=
```
anxolin, line 9: ubernit: pick we can always add some comment space things out, group them, and give more context

apps/twap/src/app/plugins/orderbook.ts
```diff
@@ -0,0 +1,106 @@
...
+  getExecutionInfo: ReturnType<typeof getExecutionInfoFactory>;
+}
+
+declare module 'fastify' {
+  interface FastifyInstance {
+    orderbook: {
+      goerli: Orderbook;
+      mainnet: Orderbook;
+      [SupportedChainId.GOERLI]: Orderbook;
+      [SupportedChainId.MAINNET]: Orderbook;
```
anxolin, line 103: why do we add 2 times the orderbook? is it one added by its number an done by its name?

apps/twap/src/app/plugins/env.ts
```diff
@@ -26,6 +30,18 @@ const schema = {
     DATABASE_PASSWORD: {
       type: 'string',
     },
+    ORDERBOOK_DATABASE_HOST: {
```
anxolin, line 33: If there's 2 databases, shou't both have a prefix name? I mean, if `ORDERBOOK_DATABASE_HOST` shouldn't we name `DATABASE_NAME` --> `XXXXX_DATABASE_NAME`?

anxolin (commented):
Great PR! 
I really like this https://typeorm.io


Regarding the salt was already in the code, so no part of this PR. Should the salt be hardcoded? doesn't it need to be unique? 
https://github.com/cowprotocol/cow-web-services/blob/feature/orderbook-integration/apps/twap/src/app/utils/getConditionalOrderId.ts#L73

PR #8, opened 2023-07-14
feat(onchain-indexing): ✨ add indexing on cold start scenarios

tukantje:
## Warning
Please first review https://github.com/cowprotocol/cow-web-services/pull/7. This PR's size will be reduced once that is merged, as this branch is made off of `feature/orderbook` branch.

For actual diff, you can also look at https://github.com/cowprotocol/cow-web-services/compare/feature/orderbook-integration...feature/onchain-indexing

## What is this?
This PR adds on chain event indexing for TWAP orders. Changes here can be categorised as;
### utils
These are various utils to derive state out of ConditionalOrder data. In general, referring to cowswap project is useful, as they are derived from there.
- calculateValidTo: returns how long an order is valid, same as https://github.com/cowprotocol/cowswap/blob/develop/src/modules/twap/updaters/PartOrdersUpdater.tsx#L103
- computeOrderUid: returns order uid, used for generating order uid for order parts, same as https://github.com/cowprotocol/cowswap/blob/develop/src/utils/orderUtils/computeOrderUid.ts#L4
- generateOrderParts: returns/derives order parts for a given twap order, similar to https://github.com/cowprotocol/cowswap/commit/7901e1ac0fa6df6410be78382949675ac94be8a5
- getApiBaseUrl: return api base url for CoW api.
- getConditionalOrderId: derives conditional order id (order id of a twap), it has to be in sync with FE to be able to derive the same id for same order, same as https://github.com/cowprotocol/cowswap/blob/develop/src/modules/twap/utils/getConditionalOrderId.ts#L8
- getOrderStatus: derives order status, mainly;
  - if amount sold is same as total sell amount we expect (partSellAmount * n) => Fulfilled
  - if `singleOrder` is false => Cancelled. This is an array filled with booleans, as per https://github.com/rndlabs/composable-cow/blob/main/src/ComposableCoW.sol#L53, that is turned to false once an on chain cancellation happens as per https://github.com/rndlabs/composable-cow/blob/main/src/ComposableCoW.sol#L143C9-L143C9
  - if ConditionalOrderCreated event's block timestamp + timeInterval * amount of parts twap order has is in the past => Expired
  - if we don't have a block timestamp (not possible in this PR as we are only dealing with mined blocks) => WaitSigning
  - everything else => Pending
- parseOrderStruct: parses staticInput field of a twap order event (`event.args.params.staticInput`), has information defining the twap order such as amount of parts, sellAmount per part etc. https://github.com/cowprotocol/cowswap/blob/develop/src/modules/twap/utils/parseTwapOrderStruct.ts#L19
- transformers: added in #7 includes transformers to marshall / unmarshall database values.

### types
Currently only has `order.ts` here, all types in application should go here.

### migrations
Database migrations, as generated & adjusted afterwards via using the command `yarn run twap:generate-migrations`. As this PR includes database migrations, please don't forget to run migrations via `yarn run twap:run-migrations` (if you don't application will run it for you on first start, however still a good practice to do it yourself to see logs easier).

### libs/abis/src/generated/custom
This is an odd change, I am not happy with it, however can't find an alternative to exposing `ConditionalOrderCreatedEvent` otherwise. If you have a better option, please feel free to suggest it. It is useful to type events we get properly.

### plugins
`infura.ts` is where the fun happens here. This plugin essentially works in the following way;
- Per network, create an infura provider
- Fetch previously processed block
- Fetch all `ConditionalOrderCreated` events from this block until current block - some amount of lag (defined per chainId, at https://github.com/cowprotocol/cow-web-services/blob/9ce7b7243d032cefeafce38b46f77c3309778240/apps/twap/src/app/plugins/infura.ts#L20)
- Process these events with previous utilities mentioned, until we have a twap order structure we can rely on, and try to save this in a transaction with the current block number. This is an upsert operation, meaning if the order exists we just end up updating its value.

### data
We have updated `order.ts` here, to include a `status` field and to add a relation to a new entity `orderPart.ts`. Order part is as it sounds, the underlying orders of a twap, so that when you are processing a twap order you know which actual orders it related to.

We have also added `block.ts` which is a table to store blockNumbers we have processed.

## How to test?
Best way to test this is via connecting to your local database to see the data in there. To avoid inflating this PR, I'll add CRUD methods in a separate PR.


apps/twap/src/app/utils/getOrderStatus.ts
```diff
@@ -0,0 +1,42 @@
+import { OrderStatus, OrderStruct } from '../types/order';
+
+function isOrderFulfilled(executedSellAmount: bigint, orderInfo: OrderStruct) {
```
shoom3301, line 3: Ubernitpick: return tyoe

apps/twap/src/app/utils/computeOrderUid.ts
```diff
@@ -0,0 +1,28 @@
...
+function isOrder(order: Order | OrderParameters): order is Order {
+  return true;
+}
+
+export async function computeOrderUid(
+  chainId: SupportedChainId,
+  owner: string,
+  order: Order | OrderParameters
+): Promise<string> {
+  const { computeOrderUid: _computeOrderUid } = await import(
```
shoom3301, line 18: Since you don't care about bundle size on backend, you can make this method synchronous

apps/twap/src/app/utils/computeOrderUid.ts
```diff
@@ -0,0 +1,28 @@
...
+    '@cowprotocol/contracts'
+  );
+  const domain = await OrderSigningUtils.getDomain(chainId);
```
shoom3301, line 21: Actually, you can make a key-value const for chainId-domain

apps/twap/src/app/data/order.ts
```diff
@@ -72,4 +77,41 @@ export class Order {
 
   @Column('int')
   span: number;
+
+  @Column('enum', {
+    enum: [
```
nenadV91, line 82: Should we move this to some const?

apps/twap/src/app/plugins/infura.ts
```diff
@@ -0,0 +1,237 @@
...
+// Add lag per network (in confirmations)
+const NETWORKS = ['mainnet', 'goerli'];
+const CONFIRMATIONS: Record<SupportedChainId, number> = {
+  '1': 5,
+  '5': 3,
+  '100': 3,
+};
+
+export const COMPOSABLE_COW_ADDRESS: Record<SupportedChainId, string> = {
+  1: '0xF487887DA5a4b4e3eC114FDAd97dc0F785d72738',
```
nenadV91, line 27: Same comment as above

apps/twap/src/app/plugins/infura.ts
```diff
@@ -0,0 +1,237 @@
...
+  salt: string;
+  handler: string;
+}
+
+export interface TwapOrdersSafeData {
+  conditionalOrderParams: ConditionalOrderParams;
+  safeTxParams: SafeTransactionParams;
+}
+
+function infuraPlugin(
```
nenadV91, line 62: This was a bit hard to follow. 
We could add some more comments explaining what is happening.

apps/twap/src/app/utils/getApiBaseUrl.ts
```diff
@@ -0,0 +1,15 @@
+import { SupportedChainId } from '@cowprotocol/cow-sdk';
+
+const COW_API_BASE_URL = 'https://api.cow.fi';
+
+const CHAIN_ID_TO_NAME: Record<SupportedChainId, string> = {
+  1: 'mainnet',
```
nenadV91, line 6: Same comment as [here](https://github.com/cowprotocol/cow-web-services/pull/8/files#r1265224178)

package.json
```diff
@@ -28,6 +28,7 @@
     "fastify": "~4.13.0",
     "fastify-cli": "^5.7.1",
     "fastify-plugin": "~4.5.0",
+    "graphql": "^16.7.1",
```
nenadV91, line 31: Is graphql used?

tukantje: This was added due to a peer dependency issue earlier - so I think one of the packages do without specifying it.

apps/twap/src/app/utils/getOrderStatus.ts
tukantje: Great one - adding it

apps/twap/src/app/utils/computeOrderUid.ts
tukantje: Good point, moved it

apps/twap/src/app/utils/computeOrderUid.ts
tukantje: Good one, adding a cache for this in proposed way (to handle async nature as well)

apps/twap/src/app/data/order.ts
tukantje: Great idea, moved it

apps/twap/src/app/plugins/infura.ts
```diff
@@ -0,0 +1,237 @@
...
+import { generateOrderParts } from '../utils/generateOrderParts';
+import { Wallet } from '../data/wallet';
+import { Order } from '../data/order';
+import { OrderPart } from '../data/orderPart';
+import { Block } from '../data/block';
+
+// Add lag per network (in confirmations)
+const NETWORKS = ['mainnet', 'goerli'];
+const CONFIRMATIONS: Record<SupportedChainId, number> = {
+  '1': 5,
```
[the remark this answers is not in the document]
tukantje: Great idea, updated it

apps/twap/src/app/plugins/infura.ts
tukantje: Great idea, simplified it

apps/twap/src/app/utils/getApiBaseUrl.ts
tukantje: Great point, updated it

apps/twap/src/app/plugins/infura.ts
tukantje: Good one, I'll update this. I should be able to extract the async function in here.

apps/twap/src/app/utils/processConditionalOrders.ts
```diff
@@ -0,0 +1,154 @@
...
+import { OrderPart } from '../data/orderPart';
+import { Block } from '../data/block';
+import { getSingleOrder } from './getSingleOrder';
+import { FastifyInstance } from 'fastify';
+import { ComposableCoW__factory } from '@cow-web-services/abis';
+import { getBlockInfo } from './getBlockInfo';
+export const COMPOSABLE_COW_ADDRESS =
+  '0xF487887DA5a4b4e3eC114FDAd97dc0F785d72738';
+
+export async function processConditionalOrders(
```
anxolin, line 17: in general, i see a lot of utils that are more doing some bussiness logic than being a util.

Also, is see bussines logic mixed with `fastify`.

However, since you were consistent and for now the project is small, you can ignore this comment. If this is fastify way of doing things, let me know!

apps/twap/src/app/utils/processConditionalOrders.ts
```diff
@@ -0,0 +1,154 @@
...
+      // An error has happened, skip event and log it.
+      console.log(
+        `Skipping:\n\tOwner: ${
+          event.args.owner
+        }\n\tOrder ID: ${getConditionalOrderId(
+          event.args.params
+        )}\n\tTx Hash: ${event.transactionHash}\n\tError: ${err.message}`
+      );
+    }
+  }
```
anxolin, line 153: i would thing this method is a bit big. Consider simplifying it

apps/twap/src/app/plugins/infura.ts
```diff
@@ -0,0 +1,64 @@
...
+}
+
+export interface TwapOrdersSafeData {
+  conditionalOrderParams: ConditionalOrderParams;
+  safeTxParams: SafeTransactionParams;
+}
+
+function infuraPlugin(
+  fastify: FastifyInstance,
+  options: Record<string, unknown>,
```
anxolin, line 41: lint warning here

anxolin (commented):
Excited about this

anxolin:
Unrelated, but should we call this project BFF since its called everywhere else like that? 

https://github.com/cowprotocol/bff?

apps/twap/src/app/utils/getBlockInfo.ts
```diff
@@ -0,0 +1,55 @@
...
+  [SupportedChainId.MAINNET]: true,
+  [SupportedChainId.GOERLI]: true,
+  [SupportedChainId.GNOSIS_CHAIN]: true,
+};
+
+// Add lag per network (in confirmations)
+const CONFIRMATIONS: Record<SupportedChainId, number> = {
+  [SupportedChainId.MAINNET]: 5,
+  [SupportedChainId.GOERLI]: 3,
+  [SupportedChainId.GNOSIS_CHAIN]: 3,
```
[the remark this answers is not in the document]
tukantje: Just to exemplify, any suggestions are welcome here on my side 😅

apps/twap/src/app/utils/getBlockInfo.ts
```diff
@@ -0,0 +1,55 @@
...
+  // We process last block again on purpose, on cold start.
+  // This way, we can handle any half - processed blocks.
+  // However, this means there will always be _at least_ 1 event processed.
+  const fromBlockNumber = lastProcessedBlock
+    ? lastProcessedBlock.blockNumber + (isColdStart ? 0 : 1)
+    : undefined;
+  isColdStart[chainId] = false;
+  const lastBlockNumber = await provider.getBlockNumber();
+  const confirmationsRequired = CONFIRMATIONS[chainId];
+  const toBlockNumber = lastBlockNumber - confirmationsRequired;
```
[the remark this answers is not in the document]
tukantje: You would get no events, in essence. Because they are named, it doesn't pose any issues to my knowledge (or from what I can tell).

apps/twap/src/app/utils/getConditionalOrderId.ts
```diff
@@ -66,11 +50,11 @@ export function twapOrderToStruct(order: TWAPOrder): TWAPOrderStruct {
 
 export function buildTwapOrderParamsStruct(
   chainId: SupportedChainId,
-  twapOrderData: TWAPOrderStruct
+  twapOrderData: OrderStruct
 ): ConditionalOrderParams {
   return {
     handler: TWAP_HANDLER_ADDRESS[chainId],
-    salt: '0x00000000000000000000000000000000000000000000000000000018920d8ce7', // hexZeroPad(Buffer.from(Date.now().toString(16), 'hex'), 32),
+    salt: '0x00000000000000000000000000000000000000000000000000000018920d8ce7',
```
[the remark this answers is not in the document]
tukantje: I should've added a comment on this as well. In essence, this salt must match between FE and BE to produce the same id, so we need to sync it.

If we use time here, because time on server vs local differ, we end up with mismatching ids.

cc/ @shoom3301

apps/twap/src/app/utils/getOrderStatus.ts
```diff
@@ -0,0 +1,45 @@
...
+  executedSellAmount: bigint,
+  orderInfo: OrderStruct,
+  blockTimestamp: Date,
+  singleOrder: boolean
+): OrderStatus {
+  if (isOrderFulfilled(executedSellAmount, orderInfo)) {
+    return OrderStatus.Fulfilled;
+  }
+
+  if (singleOrder === false) {
```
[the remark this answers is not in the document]
tukantje: Visibility, but can update it. It was written in this manner in the original file.

tukantje:
> Unrelated, but should we call this project BFF since its called everywhere else like that?
> 
> https://github.com/cowprotocol/bff?

Unifying our language might help, and CWS is quite a mouthful. So on my side it makes sense.

apps/twap/src/app/utils/parseOrderStruct.ts
```diff
@@ -0,0 +1,33 @@
...
+  { name: 't', type: 'uint256' },
+  { name: 'span', type: 'uint256' },
+  { name: 'appData', type: 'bytes32' },
+] as ParamType[];
+
+export function parseOrderStruct(staticInput: string): OrderStruct {
+  const rawData = defaultAbiCoder.decode(ABI, staticInput);
+
+  return {
+    sellToken: rawData['sellToken'],
```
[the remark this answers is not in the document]
tukantje: Type issues, you'd need to explicitly type and check rawData as well. This access pattern is TS way of circumventing that (i.e. access this field, without caring for it to be present in the type).

apps/twap/src/app/utils/processConditionalOrders.ts
tukantje: Good point, I'll categorise these as handlers in that case; since they are not utils in the strictest sense of the word indeed.

As far as the fastify instance goes, that's essentially poor man's dependency injection at the moment. Missing part here most likely is that this should be a plugin as well (fastify really likes plugins), so that I can have fastify instance accessible. Will update it tomorrow as part of a bit of a refactor here.

apps/twap/src/app/utils/processConditionalOrders.ts
tukantje: Good point, will do tomorrow.

apps/twap/src/app/utils/getBlockInfo.ts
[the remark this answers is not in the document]
anxolin: Maybe, check the average reorg size to make an inform decision of this trade-off. 
In goerli i don't care much

apps/twap/src/app/utils/calculateValidTo.ts
```diff
@@ -0,0 +1,19 @@
...
+}
+
+export function calculateValidTo({
+  part,
+  startTime,
+  frequency,
+  span,
+}: CalculateValidToParams): number {
+  if (span === 0) {
+    return startTime + (part + 1) * frequency - 1;
```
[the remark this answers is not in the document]
tukantje: Added it

apps/twap/src/app/utils/computeOrderUid.ts
```diff
@@ -0,0 +1,39 @@
...
+  domainCache[chainId] = domain;
+
+  return domain;
+}
+
+export async function computeOrderUid(
+  chainId: SupportedChainId,
+  owner: string,
+  order: Order | OrderParameters
+): Promise<string> {
```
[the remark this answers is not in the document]
tukantje: Added a TODO here, definitely a good idea

apps/twap/src/app/utils/getBlockInfo.ts
```diff
@@ -0,0 +1,55 @@
+import { SupportedChainId } from '@cowprotocol/cow-sdk';
+import { FastifyInstance } from 'fastify';
+import { Block } from '../data/block';
+import { providers } from 'ethers';
+
+// Determine the cold start per chain, to be able to process last block again
+// This avoids half-processed states.
```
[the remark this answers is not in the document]
tukantje: Added it

apps/twap/src/app/utils/getBlockInfo.ts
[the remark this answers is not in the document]
tukantje: Updated values based on Gnosis Chain's governance decisions. They use 20 blocks for finality, also for mainnet. Circle does 12, Coinbase does 14.

apps/twap/src/app/utils/getSingleOrder.ts
```diff
@@ -0,0 +1,16 @@
+import {
+  ComposableCoW,
+  ConditionalOrderCreatedEvent,
+} from '@cow-web-services/abis';
+
+// Gets the value of singleOrders mapping from ComposableCoW
+// Mainly used to determine if an order is Cancelled.
```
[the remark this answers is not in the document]
tukantje: Updated and added more to other functions.

apps/twap/src/app/utils/getBlockInfo.ts
```diff
@@ -0,0 +1,55 @@
...
+        blockNumber: 'DESC',
+      },
+      take: 1,
+    })
+  )[0];
+  // We process last block again on purpose, on cold start.
+  // This way, we can handle any half - processed blocks.
+  // However, this means there will always be _at least_ 1 event processed.
+  const fromBlockNumber = lastProcessedBlock
+    ? lastProcessedBlock.blockNumber + (isColdStart ? 0 : 1)
```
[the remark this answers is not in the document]
tukantje: Typo on my side, thank you. Fixed it.

apps/twap/src/app/utils/processConditionalOrders.ts
tukantje: Updated the structure. I've essentially;
* Created a new folder `conditionalOrders`. This would be in line with nodejs best practices (domain based folder structures, as outlined at https://github.com/goldbergyoni/nodebestpractices#-11-structure-your-solution-by-business-components)
* Moved logic that is not a util to `conditionalOrders`.
* Moved `processConditionalOrders` to plugins as a new plugin. Also created `blockInfo` plugin to simplify, and added a new function `getConditionalOrderCreatedEvents` under `conditionalOrders`.

apps/twap/src/app/utils/processConditionalOrders.ts
tukantje: Simplified this method as much as I could (details: https://github.com/cowprotocol/cow-web-services/pull/8#discussion_r1268739815).

Essentially what I could do was;
* Extract fetching of blockInfo into its own plugin
* Extract fetching of events into a separate function under `conditionalOrders`
* Transformed `processConditionalOrders` to a plugin. This way, we have `fastify` instance given to us as opposed to having it provided every time it is called.

Remaining parts are now scaffolding and massaging the results into a transaction. I don't think we'd have any more benefits from reducing size here, as we'd end up with a lot of parameter passing.

apps/twap/src/app/plugins/infura.ts
tukantje: Renamed to `_` to resolve lint warning.

apps/twap/src/app/conditionalOrder/computeOrderUid.ts
```diff
@@ -0,0 +1,48 @@
...
+  SupportedChainId,
+} from '@cowprotocol/cow-sdk';
+import { returnNever } from '../types/order';
+import { TypedDataDomain } from 'ethers';
+
+function isOrder(order: Order | OrderParameters): order is Order {
+  return true;
+}
+
+const domainCache: Partial<Record<SupportedChainId, TypedDataDomain>> = {};
```
nenadV91, line 15: If this doesn't change at all, would it make sense to cache it in a localStorage?

nenadV91: If you think it might change, ignore this comment.

apps/twap/src/app/conditionalOrder/generateOrderParts.ts
```diff
@@ -0,0 +1,74 @@
...
+        kind: OrderKind.SELL,
+        partiallyFillable: false,
+        sellTokenBalance: SellTokenSource.ERC20,
+        buyTokenBalance: BuyTokenDestination.ERC20,
+      };
+      const uid = await computeOrderUid(chainId, safeAddress, parameters);
+
+      return {
+        uid,
+        index: index,
```
nenadV91, line 66: Just `index` here?

apps/twap/src/app/conditionalOrder/getConditionalOrderId.ts
```diff
@@ -47,37 +31,35 @@ export const TWAP_HANDLER_ADDRESS: Record<SupportedChainId, string> = {
...
+ */
 export function buildTwapOrderParamsStruct(
   chainId: SupportedChainId,
-  twapOrderData: TWAPOrderStruct
+  twapOrderData: OrderStruct
 ): ConditionalOrderParams {
   return {
     handler: TWAP_HANDLER_ADDRESS[chainId],
-    salt: '0x00000000000000000000000000000000000000000000000000000018920d8ce7', // hexZeroPad(Buffer.from(Date.now().toString(16), 'hex'), 32),
+    // This salt must match between FE and BE to produce the same id, so we need to sync it.
```
nenadV91, line 47: In that case, would it be better to move it somewhere else, for example cow-sdk, and then import it?

apps/twap/src/app/conditionalOrder/getOrderStatus.ts
```diff
@@ -0,0 +1,68 @@
...
+ * @param orderInfo OrderStruct of the order.
+ * @returns If the order is fulfilled.
+ */
+function isOrderFulfilled(
+  executedSellAmount: bigint,
+  orderInfo: OrderStruct
+): boolean {
+  return (
+    executedSellAmount ===
+    BigInt(orderInfo.partSellAmount) * BigInt(orderInfo.n)
```
nenadV91, line 16: So this compares `executedSellAmount` of "parent" TWAP order with the `partSellAmount` * `number of parts`. Since each part is executed as a single order, could there be a mismatch between these two amounts even though all parts are executed. For example, can total sell amount of parts be bigger then the `executedSellAmount` due to surplus or something like that? Just asking out of curiosity.

nenadV91 (approved):
Approve with some comments.
Nice work 👍

shoom3301 (approved):

PR #31, opened 2024-05-24
Dont include line break in version

anxolin:
This PR deletes a line break that was added into the version file:

image


alfetopito (commented):
🥳

PR #32, opened 2024-05-27
Auto-deploy to staging when merging in main branch

anxolin:
This PR adds the auto-deployment to staging. 

Every time a PR is merged with `main`, it will notify the auto-deployer so it can restart the services that depend on the newly generated images.

Because of the image pull policy, the restart will pull and use the new images

## Before merging
@ahhda, I think we need to set this secrets, but I'm unsure about the value. 
~secrets.AUTODEPLOY_TAG~ (edit: not needed)
secrets.AUTODEPLOY_URL
secrets.AUTODEPLOY_TOKEN

I guess the URL is what i see in 1password in "K8S autodeploy auth", the token I assume is the "password". What is the TAG?

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -59,3 +59,13 @@ jobs:
           labels: ${{ steps.image_meta.outputs.labels }}
           builder: ${{ steps.setup-buildx.outputs.name }}
           platforms: linux/amd64
+
+
+      - uses: cowprotocol/autodeploy-action@v2
+        if: ${{ github.ref == 'refs/heads/main' }}
+        with:
+          images: ghcr.io/cowprotocol/bff/${{ env.PACKAGE_NAME }}:main
+          tag: ${{ secrets.AUTODEPLOY_TAG }}
+          url: ${{ secrets.AUTODEPLOY_URL }}
+          token: ${{ secrets.AUTODEPLOY_TOKEN }}
+          timeout: 600000 # 10 minutes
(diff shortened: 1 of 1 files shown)


ahhda:
AUTODEPLOY_TAG is not a required value when using dockerImage for restarts.

AUTODEPLOY_URL will be the ingress url for the auto deploy for the specific namespace. If you checkout the ingress list in staging cluster, you'll see the URL's. Since BFF is deployed on a separate namespace, we'll have to deploy another autodeploy for the `frontend` namespace.

.github/workflows/docker.yml
```diff
@@ -59,3 +59,13 @@ jobs:
           labels: ${{ steps.image_meta.outputs.labels }}
           builder: ${{ steps.setup-buildx.outputs.name }}
           platforms: linux/amd64
+
+
+      - uses: cowprotocol/autodeploy-action@v2
+        if: ${{ github.ref == 'refs/heads/main' }}
+        with:
+          images: ghcr.io/cowprotocol/bff/${{ env.PACKAGE_NAME }}:main
+          tag: ${{ secrets.AUTODEPLOY_TAG }}
```
ahhda, line 68:
```suggestion
```

This is not required

anxolin:
But are we perpetuating the name `frontend` for the namespace? 

Why not just use perifery? 
Is changing namespace as easy as delete the resources, change the namespace and deploy?

anxolin:
Anuj, I will merge this PR to test it out. I managed to find my way and added the secrets in the repository.

PR #33, opened 2024-05-28
feat: expose notifications api

anxolin:
This PR exposes the notifications API in BFF.

image

Right now the Nofications are coming from the CMS, but some notifications can be added to some other sources (like a database, etc), also the notifications can be enhanced with on-chain data, or post-processed, for this reason, I think is convenient to expose it through BFF.

This would also allow us to remove the dependency from CoW Swap to the CMS and reduce the permissions of reading the notifications from the public endpoint (now is a securitized request)

## Test
Run locally and execute:
`curl -i http://localhost:3010/accounts/0x79063d9173C09887d536924E2F6eADbaBAc099f5/notifications`

### DIFF
--- a/apps/api/src/app/routes/accounts/_account/notifications.ts
+++ b/apps/api/src/app/routes/accounts/_account/notifications.ts
@@ -4,6 +4,7 @@ import {
 } from '@cowprotocol/cms-api';
 import { FastifyPluginAsync } from 'fastify';
 import { FromSchema, JSONSchema } from 'json-schema-to-ts';
+import { ETHEREUM_ADDRESS_PATTERN } from '../../../schemas';
 
 const routeSchema = {
   type: 'object',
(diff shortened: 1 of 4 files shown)


apps/api/src/app/routes/accounts/_account/notifications.ts
```diff
@@ -0,0 +1,42 @@
...
+  fastify.get<{
+    Params: GetNotificationsSchema;
+    Reply: NotificationModel[];
+  }>(
+    '/notifications',
+    { schema: { params: routeSchema } },
+    async function (request, reply) {
+      const account = request.params.account;
+      const notifications = await getNotificationsByAccount({ account });
+      reply.send([]);
```
alfetopito, line 37: Shouldn't it send the notifications in the reply?

anxolin: shiiii, this was a test, it should yep

anxolin: Fixed

anxolin:
Merging to consolidate and test the release. I address issues raised after the merge

PR #35, opened 2024-06-06
Delete legacy proxy endpoint

anxolin:
BFF is live, so we don't need anymore serveless functions from Vercel.

This PR deletes the legacy proxy


alfetopito (approved):
This was never used on the self hosted BFF, so it's safe to remove anytime.

PR #36, opened 2024-06-06
Misc on improving debugability and making sure CACHEs work

anxolin:
This PR has a lot of misc changes oriented to test the correct caching of the API.
This PR will improve our debuggability, and will make it easier to test in staging some PR.

I will use it to make sure even if there was not CDN, our AWS won't ask our API twice for a resource that is cacheable. 

## Build on PRs
First of all, we want to test this in our infra (AWS), so it would be easier if we can test already before needing to merge this PR.

For this reason, I made the build of packages be triggered on every commit. This will publish new packages on every change. 

We can use this packages in staging k8s, making it easier to test

You can see now a new tag: `pr-36`: https://github.com/cowprotocol/bff/pkgs/container/bff%2Fapi

image

This allow us to run it even before this PR is merged:

image

## Logging support
Adds a new environment variable called `LOG_LEVEL`. This will allow us to control, without redeploying the log level of the app.

## Delete some additional headers
Deletes some additional headers. More specifically, it deletes Cloudflare headers, I don't think we should propagate.

It also changes how I do it. I just thought it was easier to maintain if we keep an array of the ones we want to drop.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -5,6 +5,9 @@ on:
     branches: [main]
     tags: [v*]
 
+  pull_request:
+    types: [opened, synchronize]
+
 jobs:
   docker:
     runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


anxolin:
Test failed! 

I did a bunch of requests over the course of 1min, and I saw 48 logs of the proxy call).

As it should cache over 1.5min I can verify the cache is not effective.

image

alfetopito (approved):

PR #37, opened 2024-06-17
Add redis and in-memory cache

anxolin:
This PR adds our own caching logics

## Caching
First of all, it adds a Fastify pluging `caching` to allow to easily define the cache-control headers.
It also adds in the `fastify` instance a `cache` object which allow to get/save objects from/to an abstract cache.

This allows us to not depend on any particular cache implementantion for the main caching logic.

## Redis vs In-memory
The environment var allows now to specify:
~REDIS_ENABLED=true~ (~false by default~)

EDIT: Now its enabled if you specify `REDIS_HOST` environment variable.

When Redis is enabled, the caching will happen in Redis. Otherwise, it will use an in-memory cache (not suitable for large applications in production)

## Redis integration
Additionally, I added the redis plugin, which allows us to access the redis logic through the convenient `fastify.redis` object.

This would allow us to be able to access to more advance uses of redis to implement any of the services.

TODO: Pending to not load it if `REDIS_ENABLED=false`

## Bff cache
Additionally, I've done a plugin so we can have our own logic to automatically cache things based on the cache-control directive.

It should be very easy to use now. Each endpoint will define its own caching policy using the `cache-control` directive

Let's see it with these 2 endpoints:
```ts
fastify.get('/hello', async function (request, reply) {
    await sleep(2000)
    reply.send({ hello: 'world' })
  });

  fastify.get('/hello-cached', async function (request, reply) {
    await sleep(2000)
    console.log('Setting cache header')
    reply.header(CACHE_CONTROL_HEADER, getCacheControlHeaderValue(10))
    reply.send({ hello: 'world' })
  });
```

Both endpoints return the same data after 2 seconds. The difference is `/hello-cached` returns a `cache-control` header. This header is observed by `bffCache` plugin and will cache its content for the instructed time.

Additionally, future requests it will return the cached content and will keep track of the TTL at the time of the request so it can inform the clients about it.

### Test hello endpoint
Execute several times:
`curl -i http://localhost:3010/examples/hello`

You should notice a 2seconds delay in the response every single time.

Also, there's no headers for `cache-control` or `x-bff-cache` 

### Test hello-cached endpoint
Execute this command:
`curl -i http://localhost:3010/examples/hello-cached`

Observe how after the first request, next requests are cached for 10 seconds. After that, you should experience the delay again. Additionally, the `cache-control` header should inform about the TTL so clients know when the resource is not consider fresh anymore.

image

### Credentials
Added some envs to control the redis connection

```bash
REDIS_HOST=true
REDIS_PORT=6379
REDIS_USER=redis_user
REDIS_PASSWORD=redis_user
```

## Follow ups
After this PR we still have some works to do:
- Persistance
- Schedule backup/dump of Redis
- Infrastructure setups


apps/api/src/app/plugins/bffCache.ts
```diff
@@ -0,0 +1,78 @@
...
+  ttl?: number
+}
+
+export const bffCache: FastifyPluginCallback<BffCacheOptions> = (fastify, opts, next) => {
+  const { ttl } = opts
+  fastify.addHook('onRequest', async (request, reply) => {
+    // Cache only GET requests
+    if (request.method !== 'GET') {
+      return
+    }
```
alfetopito, line 17: What if the GET request returns an error? Anything that's non 2xx

alfetopito (approved):
A question about caching failed GET requests.

Also, what about adding the a Redis instance to the docker compose for local development?

apps/api/src/app/plugins/bffCache.ts
anxolin: Good point, I will not cache those

docker-compose.yml
```diff
@@ -94,6 +94,24 @@ services:
       timeout: 10s
       retries: 5
 
+  redis:
+    container_name: redis
+    # platform: linux/amd64 -- no platform as it breaks on my mac m1
```
alfetopito, line 103: Maybe that's not needed 🤷 ?

docker-compose.yml
```diff
@@ -94,6 +94,24 @@ services:
...
+    image: redis:7.2.5
+    restart: unless-stopped
+    ports:
+      - '6379:6379'
+    volumes:
+      - data-redis:/data
+      #- ./redis.conf:/usr/local/etc/redis/redis.conf
```
alfetopito: Dead code?

apps/api/src/app/plugins/bffCache.ts
```diff
@@ -0,0 +1,80 @@
...
+      }
+    }
+
+    return
+  })
+
+  fastify.addHook('onSend', function (req, reply, payload, next) {
+    const isCacheHit = reply.getHeader(HEADER_NAME) === 'HIT'
+    const cacheTtl: number | undefined = getTtlFromResponse(reply, ttl)
+    const isStatus200 = reply.statusCode < 200 || reply.statusCode >= 300
```
alfetopito, line 45: The variable name is inverted.
You are filtering out when it's not a 200.
Either update the name or the condition:

```suggestion
    const isNotStatus200 = reply.statusCode < 200 || reply.statusCode >= 300
```

```suggestion
    const isStatus200 = reply.statusCode >= 200 && reply.statusCode < 300
```

PR #41, opened 2024-06-20
Fix overriding cache headers

anxolin:
The cache override had a bug, and was never overriding the cache.

It was always returning 30s cache, which comes from coingecko.

I used now `COINGECKO_CACHING_TIME=3600` and has effect:

image


alfetopito (commented):
👍

PR #43, opened 2024-06-28
Add default slippage endpoint

anxolin:
This PR makes a mock endpoint we can use to get the default slippage

image

This way, we can implement on a later PR our heuristic (using coingecko API, orderbook API, CMS, or whatever we see suitable)

## Validation
The endpoint will validate the chain is an integer:
image

Also that is supported chain:

image

Same for the market, it will require they are Ethereum addresses.

## Next steps
This is a mock value that returns our current default, we can discuss the heuristic now and iterate on this.


anxolin:
I do post-merge reviews

I just want this in staging

PR #48, opened 2024-07-08
Add slippage service - suggest 3 layered architecture

anxolin:
This PR is the first of a series I'd like to add to BFF.

## Context on why a service library
We will have a growing complexity in BFF and I thought about making a suggestion.

My general plan is to:
- Create 2 new libraries `services` (in this PR) and `repositories` (follow up)
- Our API will just delegate all work to a service, this way, the API layer remains lightweight and all business logic is moved to the service.
- The service might make use of one or more repositories to access the information 
- The repositories allows us to abstract how we access the data (separation on data access from the service logic)
- I would also like to work with Typescript interfaces for these 2 layers (to hide implementation details), and potentially have more than one implementation. This will help with unit testing for example, or implement a cached version of a repository.
- Finally, I would like to introduce a IoC framework (to decouple everything)

I will do the PoC just to implement some basic slippage service

## What if we need to create too many services or repositories?
For now, one library for all services and one library for all repositories suffice in my opinion. If at some point in the future, it becomes too big, we can always split it into multiple service libraries (for example, by domain).

## This PR
- [x] Adds just the service layer (lib)
- [x] Adds the slippage service 
   - [x] interface
   - [x] the implementation
   - [x] factory method which will be removed once the IoC is added

## Test 

`nx run services:test --watch`

Screenshot at Jul 08 20-55-00

### DIFF
--- a/apps/api/src/app/routes/chains/__chainId/markets/__sellToken-__buyToken/slippageTolerance.ts
+++ b/apps/api/src/app/routes/chains/__chainId/markets/__sellToken-__buyToken/slippageTolerance.ts
@@ -4,6 +4,10 @@ import {
 } from '../../../../../schemas';
 import { FastifyPluginAsync } from 'fastify';
 import { FromSchema, JSONSchema } from 'json-schema-to-ts';
+import { getSlippageService } from '@cowprotocol/services';
+
+// TODO:  Add this in a follow up PR
+// import { ALL_SUPPORTED_CHAIN_IDS } from '@cowprotocol/cow-sdk';
 
 interface Result {
   slippageBps: number;
@@ -38,7 +42,7 @@ const responseSchema = {
       description:
         'Slippage tolerance in basis points. One basis point is equivalent to 0.01% (1/100th of a percent)',
       type: 'number',
-      examples: [50, 100, 200],
+      examples: [50, 100, 200], // [ALL_SUPPORTED_CHAIN_IDS],
       minimum: 0,
       maximum: 10000,
     },
@@ -46,6 +50,8 @@ const responseSchema = {
 } as const satisfies JSONSchema;
 
 type RouteSchema = FromSchema<typeof routeSchema>;
+
+const slippageService = getSlippageService();
 
 const root: FastifyPluginAsync = async (fastify): Promise<void> => {
   // example: http://localhost:3010/chains/1/markets/0x6b175474e89094c44da98b954eedeac495271d0f-0x2260fac5e5542a773aa44fbcfedf7c193bc2c599/slippageTolerance
@@ -67,7 +73,11 @@ const root: FastifyPluginAsync = async (fastify): Promise<void> => {
       fastify.log.info(
         `Get default slippage for market ${baseTokenAddress}-${quoteTokenAddress} on chain ${chainId}`
       );
-      reply.send({ slippageBps: 50 });
+      const slippageBps = slippageService.getSlippageBps(
+        baseTokenAddress,
+        quoteTokenAddress
+      );
+      reply.send({ slippageBps });
     }
   );
 };
--- a/libs/services/src/SlippageService/SlippageService.spec.ts
+++ b/libs/services/src/SlippageService/SlippageService.spec.ts
@@ -1,0 +1,9 @@
+import { getSlippageService } from './SlippageService';
+
+const slippageService = getSlippageService();
+
+describe('SlippageService', () => {
+  it('should return always 50', () => {
+    expect(slippageService.getSlippageBps('0x0', '0x0')).toEqual(50);
+  });
+});
--- a/libs/services/src/SlippageService/SlippageService.ts
+++ b/libs/services/src/SlippageService/SlippageService.ts
@@ -1,0 +1,20 @@
+/**
+ * BPS (Basis Points)
+ */
+export type Bps = number;
+
+export interface SlippageService {
+  getSlippageBps(quoteTokenAddress: string, baseTokenAddress: string): Bps;
+}
+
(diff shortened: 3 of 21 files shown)


apps/api/src/app/routes/chains/__chainId/markets/__sellToken-__buyToken/defaultSlippageTolerance.ts
```diff
@@ -4,6 +4,10 @@ import {
 } from '../../../../../schemas';
 import { FastifyPluginAsync } from 'fastify';
 import { FromSchema, JSONSchema } from 'json-schema-to-ts';
+import { getSlippageService } from '@cowprotocol/services';
+
+// TODO:  Add this in a follow up PR
+// import { ALL_SUPPORTED_CHAIN_IDS } from '@cowprotocol/cow-sdk';
```
anxolin, line 10: Not for this PR, but leaving here so I don't forget :)

libs/services/src/SlippageService/SlippageService.ts
```diff
@@ -0,0 +1,20 @@
...
+}
+
+// TODO: Find good name for the implementation, as Leandro don't like "Impl" suffix, just don't want to couple it to add Coingecko in its name (as that would couple it to the data-source, and the adding a name to specify the algorithm might complicate the name, as this will use some custom logic based on standard deviation, so for now as we plan to have just one implementation, I want to keep it simple. But happy to get ideas here)
+export class SlippageServiceImpl implements SlippageService {
+  getSlippageBps(_quoteTokenAddress: string, _baseTokenAddress: string): Bps {
+    return 50;
+  }
+}
+
+// TODO: This is just temporal! I will introduce a IoC framework in a follow up
```
anxolin, line 17: Temporal factory method. this will disappear in a follow up

libs/services/src/SlippageService/SlippageService.ts
```diff
@@ -0,0 +1,20 @@
...
+export type Bps = number;
+
+export interface SlippageService {
+  getSlippageBps(quoteTokenAddress: string, baseTokenAddress: string): Bps;
+}
+
+// TODO: Find good name for the implementation, as Leandro don't like "Impl" suffix, just don't want to couple it to add Coingecko in its name (as that would couple it to the data-source, and the adding a name to specify the algorithm might complicate the name, as this will use some custom logic based on standard deviation, so for now as we plan to have just one implementation, I want to keep it simple. But happy to get ideas here)
+export class SlippageServiceImpl implements SlippageService {
+  getSlippageBps(_quoteTokenAddress: string, _baseTokenAddress: string): Bps {
+    return 50;
```
anxolin, line 13: Before I get into the actual implementation, I want to setup the 3 layers and the IoC. 

This way, I can abstract where do we get the USD estimations from, and we can test the algorithm of calculating the slippage independently of where do we get the USD estimation from (actually, for the `UsdRepository` test will be mocked)

libs/services/src/SlippageService/SlippageService.spec.ts
```diff
@@ -0,0 +1,9 @@
+import { getSlippageService } from './SlippageService';
+
+const slippageService = getSlippageService();
+
+describe('SlippageService', () => {
+  it('should return always 50', () => {
+    expect(slippageService.getSlippageBps('0x0', '0x0')).toEqual(50);
```
anxolin, line 7: Unit test for the service

apps/api/src/app/routes/chains/__chainId/markets/__sellToken-__buyToken/slippageTolerance.ts
```diff
@@ -38,7 +42,7 @@ const responseSchema = {
       description:
         'Slippage tolerance in basis points. One basis point is equivalent to 0.01% (1/100th of a percent)',
       type: 'number',
-      examples: [50, 100, 200],
+      examples: [50, 100, 200], // [ALL_SUPPORTED_CHAIN_IDS],
```
alfetopito, line 45: This comment doesn't match the property.

libs/services/src/SlippageService/SlippageService.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * BPS (Basis Points)
+ */
+export type Bps = number;
+
+export interface SlippageService {
+  getSlippageBps(quoteTokenAddress: string, baseTokenAddress: string): Bps;
+}
+
+// TODO: Find good name for the implementation, as Leandro don't like "Impl" suffix, just don't want to couple it to add Coingecko in its name (as that would couple it to the data-source, and the adding a name to specify the algorithm might complicate the name, as this will use some custom logic based on standard deviation, so for now as we plan to have just one implementation, I want to keep it simple. But happy to get ideas here)
```
alfetopito, line 10: ;)

alfetopito: What about: `BasicSlippageService`
Or, `StandardSlippageService`, `DefaultSlippageService`, `NaiveSlippageService` or even `FixedSlippageService` or `DumbSlippageService` at it doesn't calculate anything (yet) :P

alfetopito (approved):
Left some suggestions for the implementation class name

libs/services/src/SlippageService/SlippageService.ts
anxolin: You like it more as Prefix? 

The reason I usually add it a s sufix is to use it as a qualifier. Like, `SlippageServiceA`  and  `SlippageServiceB`  rather than `ASlippageService` and `BSlippageService`  

This way we go from the general "SlippageService" to the specific `SlippageServiceStrategyX`
Makes it easier to search for it in the EDI in my opinion. You type `SlippageService` and sorted you will see
- `SlippageServiceStrategyX`
- `SlippageServiceStrategyY`
- `SlippageServiceStrategyZ`

Regarding the actual prefix/suffix, I used `Impl` cause is short and typically there's just one implementation, and when there's more will be cause one is a `Mock` or a `Proxy` or `Redis` or some other qualifier, but most of the time I don't expect more than one implementation. 

I will call it `Main` for now as a trade off

PR #49, opened 2024-07-08
Add repository layer

anxolin:
Follow up on https://github.com/cowprotocol/bff/pull/48

Adds the repository. The Service will need a way to estimate the price of a token for a given date. 

This PR adds the interface and some mock implementation. 
When we start doing the actual implementation I would like to:
- Implement the coingecko implementation
- Make a Redis version, delegates to coingecko implementation, but uses Redis for caching

For now, just a mock :) 

## Context
https://github.com/cowprotocol/bff/pull/48

### DIFF
--- a/libs/repositories/src/UsdRepository.ts
+++ b/libs/repositories/src/UsdRepository.ts
@@ -1,0 +1,16 @@
+import { get } from 'http';
+
+export interface UsdRepository {
+  getPriceDayUsd(tokenAddress: string, date: Date): Promise<number>;
+}
+
+export class UsdRepositoryMock {
+  async getPriceDayUsd(tokenAddress: string, date: Date): Promise<number> {
+    return 1234;
+  }
+}
+
+// TODO: Remove once we have IoC
+export function getUsdRepository(): UsdRepository {
+  return new UsdRepositoryMock();
+}
--- a/libs/repositories/src/UsdRepository.ts
+++ b/libs/repositories/src/UsdRepository.ts
@@ -1,11 +1,11 @@
 import { get } from 'http';
 
 export interface UsdRepository {
-  getPriceDayUsd(tokenAddress: string, date: Date): Promise<number>;
+  getDailyUsdPrice(tokenAddress: string, date: Date): Promise<number>;
 }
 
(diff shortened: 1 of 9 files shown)


libs/repositories/src/UsdRepository.ts
```diff
@@ -0,0 +1,16 @@
+import { get } from 'http';
+
+export interface UsdRepository {
+  getPriceDayUsd(tokenAddress: string, date: Date): Promise<number>;
```
anxolin, line 4: main signature I'd like to suggest

alfetopito: It sounds weird. What about `getDailyUsdPrice`

libs/repositories/src/UsdRepository.ts
```diff
@@ -0,0 +1,16 @@
+import { get } from 'http';
+
+export interface UsdRepository {
```
alfetopito, line 3: Not sure how generic we want to go, and if we'll ever get more currencies, but this could also be modelled like:
CurrencyRepository or FiatRepository which then contains UsdRepository, EurRepository and so on.

Or it's a single CurrencyRepository or FiatRepository in which the actual currency is one of the parameters.

libs/repositories/src/UsdRepository.ts
```diff
@@ -0,0 +1,16 @@
...
+  getPriceDayUsd(tokenAddress: string, date: Date): Promise<number>;
+}
+
+export class UsdRepositoryMock {
```
alfetopito, line 7: Nitpick: why doesn't the mock implement the interface? =P

alfetopito (approved):

libs/repositories/src/UsdRepository.ts
anxolin: I though about this, and because we don't plan to change this in the near feature, I think is fine. If it changes scope, its easy to increase the scope of the repository later.

libs/repositories/src/UsdRepository.ts
anxolin: It should, is not a nit is a good catch :)

PR #51, opened 2024-07-09
Add nest framework and integrate with fastify

anxolin:
I just gave it a try to integrate `nest` and `fastify` into the `NX` project.

I managed to boot the app with the 2 servers:
- I run the good old `Fastify` API in http://localhost:3010 
- Then I run a second `Nest API  (with Fastify integration in )` http://localhost:3011

The problem is that the injection of dependencies didn't work for me.

- I made a service, that should be injected in the controller.
- I can see how the controller is created, and so is the service... but the service, which should be injected into the controller constructor is not being passed

Here you can see all the stuff I mentioned above in the logs:
Screenshot at Jul 09 21-32-17

But if you go to  http://localhost:3011, you will see:

Screenshot at Jul 09 21-32-43

The controller managed the request, but the service was `undefined`, and this is why you see the message above:

Screenshot at Jul 09 21-33-16

If I try to debug the constructor call:
image

I see how I'm being injected an undefined instead of my service:
image


anxolin:
It can be brought to live any time. Closing for now

PR #53, opened 2024-07-11
Implement an USD repository fallback

anxolin:
This PR implements a repository fallback.

The repository fallback is configured with an array of repositories and will delegate in order of specification the request to the repositories.

I created some tests that asserts it works as expected


libs/repositories/src/UsdRepository/UsdRepositoryCoingecko.test.ts
```diff
@@ -1,13 +1,12 @@
...
 import ms from 'ms';
 
-const FIVE_MINUTES = ms`5m`;
-const ONE_HOUR = ms`1h`;
-const ONE_DAY = ms`1d`;
-const WETH = '0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2';
-const DEFINITELY_NOT_A_TOKEN = '0x0000000000000000000000000000000000000000';
+const FIVE_MINUTES = ms('5m');
+const ONE_HOUR = ms('1h');
+const ONE_DAY = ms('1d');
```
alfetopito, line 9: Why change? Does not work to pre-compile the constant in this context or something?

libs/repositories/src/UsdRepository/UsdRepositoryFallback.ts
```diff
@@ -0,0 +1,38 @@
...
+  async getUsdPrice(
+    chainId: number,
+    tokenAddress: string
+  ): Promise<number | null> {
+    for (const usdRepository of this.usdRepositories) {
+      const price = await usdRepository.getUsdPrice(chainId, tokenAddress);
+      if (price !== null) {
+        return price;
+      }
+    }
```
alfetopito, line 17: Should we have a way to avoid hammering repositories that fail, for example due to rate limiting and retrying it later?
A retry strategy.

alfetopito (approved):

libs/repositories/src/UsdRepository/UsdRepositoryCoingecko.test.ts
anxolin: cause suddenly I started to get errors.... don't ask me why 😢
I just wanted the errors to go, and this worked (this is what the official library docs told me to do)

libs/repositories/src/UsdRepository/UsdRepositoryFallback.ts
anxolin: Good point, but I would not do it for the rety. I would do it more for the case where we get a `null`, which is most likely a (I don't know this token)

I could iterate, but let me first implement a second repository for getting the USD from our orderbook API

anxolin: https://github.com/cowprotocol/bff/issues/54

PR #55, opened 2024-07-11
Add orderbook API

anxolin:
This PR just adds a new client for our orderbook API, it will become handy to implement repositories

Please, for this PR ignore I didn't add the endpoints per-network. Its coming in next


libs/repositories/src/cowApi.ts
```diff
@@ -0,0 +1,9 @@
+import createClient from 'openapi-fetch';
+
+const COW_API_BASE_URL = process.env.COW_API_BASE_URL || 'https://api.cow.fi';
+
+import type { paths } from './gen/cow/cow-api-types';
+
+export const cowApiClient = createClient<paths>({
+  baseUrl: COW_API_BASE_URL,
+});
```
alfetopito, line 9: How would it look like for different chains? Do we need on client instance for each chain AND env (barn/prod), since they are all different urls?

alfetopito (approved):

libs/repositories/src/cowApi.ts
anxolin: Yes, I think I'll probably lazy instantiate a client per network (although eager also works cause is pretty lightweight and is backend code)

PR #56, opened 2024-07-16
Create cow repository for USD estimation based on native price (+tests)

anxolin:
This PR will calculate the USD price for any token by using native price endpoint of CoW API

This PR includes the implementation of the logic (basically gets native price of USDC and the token and compare it to each other)

The tests asserts that it works for some representative case: WETH (because it has 18 decimals, and USDC has 6, so we make sure it takes decimals into account.

It will test 404 case when you pass something is not a token. --> will return null

It will test 400 case
* Will test for example when the tokens are not supported --> we return null
* Other 400 errors will be throwing (I tested this too). It will try to use the description and error type returned by the backend

It will also test that any other error that is not a 200, then it will throw --> tests the 418 I'm not a teapot code

image

## Extra
This PR also includes
- Improves the setup of IoC (now all apps and libs will set it up as part of the jest config). They will all use a shared setup script
- Adds 50% error tolerance for the coingecko integration test (it was failing at some point)


libs/repositories/src/UsdRepository/UsdRepository.ts
```diff
@@ -20,11 +22,30 @@ export interface PricePoint {
...
 
   getUsdPrices(
-    chainId: number,
+    chainId: SupportedChainId,
     tokenAddress: string,
     priceStrategy: PriceStrategy
   ): Promise<PricePoint[] | null>;
 }
+
+export class UsdRepositoryNoop implements UsdRepository {
```
alfetopito, line 37: Why a noop repository?

libs/repositories/src/const.ts
```diff
@@ -0,0 +1,30 @@
...
+  decimals: number;
+}
+
+export const USDC: Record<SupportedChainId, TokenAddressAndDecimals> = {
+  [SupportedChainId.MAINNET]: {
+    address: '0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48',
+    decimals: 6,
+  },
+  [SupportedChainId.GNOSIS_CHAIN]: {
+    address: '0x2a22f9c3b484c3629090feed35f17ff8f88f76f0', // '0xDDAfbb505ad214D7b80b1f830fcCc89B60fb7A83',
```
alfetopito, line 15: Why the second commented address?

libs/repositories/src/cowApi.ts
```diff
@@ -4,6 +4,6 @@ const COW_API_BASE_URL = process.env.COW_API_BASE_URL || 'https://api.cow.fi';
 
 import type { paths } from './gen/cow/cow-api-types';
 
-export const cowApiClient = createClient<paths>({
-  baseUrl: COW_API_BASE_URL,
+export const cowApiClientMainnet = createClient<paths>({
+  baseUrl: COW_API_BASE_URL + '/mainnet',
```
alfetopito, line 8: Not for now but, once we do add the SDK, use the orderbook api instance instead?

libs/repositories/src/utils/fetch.ts
```diff
@@ -0,0 +1,14 @@
+export async function throwIfUnsuccessful(
```
alfetopito, line 1: File name doesn't match export

libs/repositories/src/UsdRepository/UsdRepository.ts
anxolin: Its convenient. Both methods return null if they don't know how to get the data. For example, its an unknown token for coingecko.

This is just convenient for extending it and only implement the getUsdPrice for the cow native price implementation, but leave the other method returning always `null`

libs/repositories/src/const.ts
anxolin: One is USDe and the one in the comment USD. I will delete the comment

libs/repositories/src/cowApi.ts
anxolin: I'd love to make the SDK to export this client. I don't love the current implementation of the SDK

libs/repositories/src/utils/fetch.ts
anxolin: `fetch` was just meant to be having a bunch of utils for fetch. For now only one. Anyways, I can rename the file if you prefer

libs/repositories/src/UsdRepository/UsdRepositoryCoingecko.ts
```diff
@@ -49,20 +48,23 @@ export class UsdRepositoryCoingecko implements UsdRepository {
...
     );
 
-    if (fetchResponse.error) {
-      throw fetchResponse.error;
-    }
-    const priceData = fetchResponse.data;
-
-    if (!priceData[tokenAddressLower] || !priceData[tokenAddressLower].usd) {
+    if (
+      response.status === 404 ||
```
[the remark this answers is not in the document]
alfetopito: What's the behaviour when it throws? 500 or return the same 4xx error?

libs/repositories/src/UsdRepository/UsdRepository.ts
alfetopito: But, returning null in some cases is not the same as a noop operation, IMO.

Maybe I'm not familiar with the class noop pattern, or it's a matter of naming. Don't what to name this, though.

libs/repositories/src/utils/fetch.ts
alfetopito: I would prefer, since looking at it seems like a mistake.
Or explain it in a comment/docstring like you just told me.

anxolin: no worries, I did the change when I wrote this comment

libs/repositories/src/const.ts
anxolin: I removed it, i don't think we need that there

libs/repositories/src/UsdRepository/UsdRepository.ts
anxolin: Because `null` means here, "I don't know", `noop" is how a repository that complies with the interface but it never know how to get USD prices, or USD price points

For me, it was just convenient primitive to have for testing and for defining repost with a "Default implementation" which just replies "I don't know" (but can be overrides)

anyways, if this confuses you, I will delete it so we don't have to give it a name, and implement the "I don't know" methods splicitly wherever I need them

libs/repositories/src/UsdRepository/UsdRepositoryCoingecko.ts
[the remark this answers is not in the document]
anxolin: The behaviour of 500 for me is similar to the any unhandled 400, I just want to throw, because is not the 200 I'm after (or the other 2 cases where I return a `null` stating I don't know the price)

In other words. I handle all expected cases, the unexpected, they throw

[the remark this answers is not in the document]
alfetopito: It's still not clear to me.
Would a 400 response from coingecko result in a 500 response from bff?

alfetopito (approved):

libs/repositories/src/UsdRepository/UsdRepositoryCoingecko.ts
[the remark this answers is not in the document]
anxolin: > Would a 400 response from coingecko result in a 500 response from bff?

It should, yes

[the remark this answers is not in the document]
alfetopito: A little, yes.
Image we have an alert for high rate of 500s that will wake someone up to deal with it.
The errors though are coming from a client (potentially not us) that's simply using the wrong parameters, or we ourselves called the upstream endpoint too often, causing a 429.

Should the service still return 500s? What would be the action oncall can take to alleviate the 500s?

I raise this mostly because 5xx errors are in my mind something unexpected that should not be happening such as a server crash. A 4xx error from an upstream service is just expected and the service is still running fine.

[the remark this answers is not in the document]
anxolin: I'll open an issue so we can keep track of this, cause I think this needs some alignment, although I see improvements on capturing more cases from falling upstream, I do think that us passing bad params to an API we delegate is an internal error on our side, not something the client did wrong.

I do think this should trigger this 500 alert in case we messed up the params, or the underling delegation service changed the params. 40X would mean its the user's fault, but 50X means its our fault. 

Then, its possible we would need to refine further and catch some cases and handle it differently. We might want to decide what to do in case of a 429. Again, this is not necessarily users's fault (it could be we have multiple concurrent users and this triggers the error, so again the issue is in our side, nothing the client can fix). The refinement could be to add some cool down and re-attempt, throttle requests so this doesn't happen, or pay for a better API Key that allows more request/second. 

Anyways, let's discuss it in an issue or personally to get an alignment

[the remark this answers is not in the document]
anxolin: https://github.com/cowprotocol/bff/issues/69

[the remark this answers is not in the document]
alfetopito: Ok, fair points

PR #60, opened 2024-07-18
Implement and expose USD service

anxolin:
## Implement USD Service
This PR implements a very basic UsdService, which just delegates to the repository. Because the service logic is dumb I didn't do any unit tests (it feels overkill, the relevant tests for USD are in the repositories)

The repository is injected, so the service doesn't care about the specific implementation.

Screenshot at Jul 18 22-40-38

Screenshot at Jul 18 22-41-47

## Expose USD API
Exposes the service through a basic API

`/chains/1/tokens/0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48/usdPrice` 

#alfetopito I know you don't like. I'm happy to do a refactor to `/1/tokens/0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48/usdPrice` but we will need to then make it backward compatible with the slippage endpoint. I added chains to be more explicit (I think is not the purest API definition, if the ID there doesn't have context), but we can compromise that for simplicity's sake. I'm mostly indifferent (slightly favour explicit, but I don't care a lot)

## Test

Run locally for a few tokens. In this PR you will notice some small delay (because there's no caching, and checks in Coingecko all the time). There's no fallback logics and native_price in COW is not used. All that, coming in follow up PRs :) 

Check different tokens in:
http://localhost:3010/chains/1/tokens/0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48/usdPrice

### DIFF
--- a/apps/api/src/app/routes/chains/__chainId/markets/__baseTokenAddress-__quoteTokenAddress/slippageTolerance.ts
+++ b/apps/api/src/app/routes/chains/__chainId/markets/__baseTokenAddress-__quoteTokenAddress/slippageTolerance.ts
@@ -11,16 +11,15 @@ import {
   getCacheControlHeaderValue,
 } from '../../../../../../utils/cache';
 
+const CACHE_SECONDS = 120;
+
 // TODO:  Add this in a follow up PR
 // import { ALL_SUPPORTED_CHAIN_IDS } from '@cowprotocol/cow-sdk';
-
-interface Result {
-  slippageBps: number;
-}
 
 const routeSchema = {
   type: 'object',
   required: ['chainId', 'baseTokenAddress', 'quoteTokenAddress'],
+  additionalProperties: false,
   properties: {
     chainId: ChainIdSchema,
     baseTokenAddress: {
@@ -38,16 +37,17 @@ const routeSchema = {
   },
 } as const satisfies JSONSchema;
 
-const responseSchema = {
+const successSchema = {
   type: 'object',
   required: ['slippageBps'],
+  additionalProperties: false,
   properties: {
     slippageBps: {
       title: 'Slippage tolerance in basis points',
       description:
         'Slippage tolerance in basis points. One basis point is equivalent to 0.01% (1/100th of a percent)',
       type: 'number',
-      examples: [50, 100, 200], // [ALL_SUPPORTED_CHAIN_IDS],
+      examples: [50, 100, 200],
       minimum: 0,
       maximum: 10000,
     },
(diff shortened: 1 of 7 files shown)


apps/api/src/app/routes/chains/__chainId/markets/__baseTokenAddress-__quoteTokenAddress/slippageTolerance.ts
```diff
@@ -11,16 +11,15 @@ import {
...
 // import { ALL_SUPPORTED_CHAIN_IDS } from '@cowprotocol/cow-sdk';
 
-interface Result {
-  slippageBps: number;
-}
-
 const routeSchema = {
   type: 'object',
   required: ['chainId', 'baseTokenAddress', 'quoteTokenAddress'],
+  additionalProperties: false,
```
alfetopito, line 22: With this the schema now accepts unknown properties.
What kind of properties do you expect?

apps/api/src/app/routes/chains/__chainId/tokens/__tokenAddress/usdPrice.ts
```diff
@@ -0,0 +1,99 @@
...
+      },
+    },
+    async function (request, reply) {
+      const { chainId, tokenAddress } = request.params;
+
+      const price = await usdService.getUsdPrice(chainId, tokenAddress);
+      fastify.log.info(
+        `Get USD value for ${tokenAddress} on chain ${chainId}: ${price}`
+      );
+      if (!price) {
```
alfetopito, line 89: Unlikely, but could it be that we get a valid price of `0` and want to return that instead of 404?

alfetopito (approved):
> #alfetopito I know you don't like. 

Read my mind :)

> but we will need to then make it backward compatible with the slippage endpoint

And that's why I wanted to change it before 🤷 

Anyway, won't be a big deal if we do, because:
1. No crucial services depend on this
2. Being down won't cause CoW Swap to stop working
3. The earlier the better

anxolin:
My next PR will remove the `chains` then :)

apps/api/src/app/routes/chains/__chainId/markets/__baseTokenAddress-__quoteTokenAddress/slippageTolerance.ts
anxolin: just the opposite, I don't accept them. I did this, to make sure the inferred type was what I was expecting

apps/api/src/app/routes/chains/__chainId/tokens/__tokenAddress/usdPrice.ts
anxolin: Good point, I will address

anxolin: I address it, but end up in the wrong PR https://github.com/cowprotocol/bff/pull/61/commits/d31d620513c5310377b25ca598b676226da4b0c4

PR #67, opened 2024-07-25
chore: hotfix v0.6.4

alfetopito:
# Summary

Changes based off of https://github.com/cowprotocol/bff/releases/tag/v0.6.3 which is the current prod version.

Pointing to the new branch `v0.6`.

It fixes the issue with cached responses.

The requests are being cached, but the cached part is the stream, not the the actual payload.
Since the stream is consumed in the first reply, the subsequent tries in the cached version return nothing:

```
"err":{"type":"FastifyError","message":"Reply was already sent.","stack":"FastifyError: Reply was already sent.\n    at _Reply.Reply.send (/Users/alfeto/work/cow/bff/node_modules/fastify/lib/reply.js:127:26)\n    at /Users/alfeto/work/cow/bff/node_modules/@fastify/reply-from/index.js:184:14\n    at /Users/alfeto/work/cow/bff/node_modules/@fastify/reply-from/index.js:277:9\n    at /Users/alfeto/work/cow/bff/node_modules/@fastify/reply-from/lib/request.js:173:7\n    at RequestHandler.runInAsyncScope (node:async_hooks:206:9)\n    at RequestHandler.onHeaders (/Users/alfeto/work/cow/bff/node_modules/undici/lib/api/api-request.js:105:14)\n    at Request.onHeaders (/Users/alfeto/work/cow/bff/node_modules/undici/lib/core/request.js:233:27)\n    at Parser.onHeadersComplete (/Users/alfeto/work/cow/bff/node_modules/undici/lib/client.js:794:23)\n    at wasm_on_headers_complete (/Users/alfeto/work/cow/bff/node_modules/undici/lib/client.js:408:30)\n    at null.<anonymous> (wasm://wasm/00036ac6:1:1173)","code":"FST_ERR_REP_ALREADY_SENT","name":"FastifyError","statusCode":500},"msg":"Reply already sent"}
```

Note: in order to properly consume and cache the payload, I had to disable the compression, similar to what I did in my first attempt at this https://github.com/cowprotocol/bff/pull/10/commits/55387a784d0698f45068027ab0ab565e3b069306

PR #68, opened 2024-07-25
fix: use url as cache key which contains query params

alfetopito:
# Summary

Now that the caching was fixed on https://github.com/cowprotocol/bff/pull/67, the next issue was exposed.

Unfortunately I only noticed after it was already deployed to prod :(

Right now, the cache is the SAME for all requests.
That's because the cache key used in the request path: `req.routerPath` which is `/proxies/coingecko/*` for all requests.

Changed to `req.url` which does contain the full url path including the search params. E.g.: `/proxies/coingecko/simple/token_price/ethereum?contract_addresses=0x6B175474E89094C44Da98b954EedeAC495271d0F&vs_currencies=usd`

# Testing

1. Start the server locally with `yarn start`
2. Query one address: `curl -X GET http://localhost:3000/proxies/coingecko/simple/token_price/ethereum?contract_addresses=0x6B175474E89094C44Da98b954EedeAC495271d0F&vs_currencies=usd`
* Should receive the price for the asset
3. Query the same asset again
* Should receive the same response
4. Query a different asset: `curl -X GET http://localhost:3000/proxies/coingecko/simple/token_price/ethereum?contract_addresses=0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2&vs_currencies=usd`
* Should have a DIFFERENT price and the token address in the response should match the one in the query

image

### DIFF
--- a/apps/api/src/app/plugins/bffCache.ts
+++ b/apps/api/src/app/plugins/bffCache.ts
@@ -17,7 +17,7 @@ export const bffCache: FastifyPluginCallback<BffCacheOptions> = (fastify, opts, 
       return
     }
 
-    let key = getKey(request)
+    const key = getKey(request);
     // Remove it so we can cache it properly
     request.headers['accept-encoding'] = undefined;
 
@@ -59,7 +59,6 @@ export const bffCache: FastifyPluginCallback<BffCacheOptions> = (fastify, opts, 
       contents += chunk.toString(); // Process each chunk of data
     }
 
-
     const key = getKey(req);
     setCache(key, contents, cacheTtl, fastify).catch((e) => {
       fastify.log.error(`Error setting key ${key} from cache`, e);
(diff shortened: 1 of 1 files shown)


alfetopito:
Tested on infra with https://app.pulumi.com/cowprotocol/bff/staging/updates/183

All good, proceeding with release. Please post merge review.

apps/api/src/app/plugins/bffCache.ts
```diff
@@ -74,7 +73,7 @@ export const bffCache: FastifyPluginCallback<BffCacheOptions> = (fastify, opts,
 };
 
 function getKey(req: FastifyRequest) {
-  return `GET:${req.routerPath}`;
+  return `GET:${req.url}`;
```
anxolin, line 76: 🙈

Thanks

PR #71, opened 2024-07-29
Increase caching time

anxolin:
This PR increases the caching time for values to 2min.

It also makes the specification of the caching time mandatory, and moves its const to the `inversify` configuration. 

## Test

To test, you can add some logs (you can apply this changes https://github.com/cowprotocol/bff/pull/71/commits/6aedf885ae3657bac8890dc6a12138dd3df7d833)

Then you can query the price of these tokens:

- curl -i  http://127.0.0.1:3010/1/tokens/0x6b175474e89094c44da98b954eedeac495271d0f/usdPrice
- curl -i  http://127.0.0.1:3010/1/tokens/0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2/usdPrice 

It might be handy to change the caching time to `10 seconds`  https://github.com/cowprotocol/bff/blob/increase-caching-time/apps/api/src/app/inversify.config.ts#L15

image

### DIFF
--- a/apps/api/src/app/inversify.config.ts
+++ b/apps/api/src/app/inversify.config.ts
@@ -12,6 +12,9 @@ import {
   usdRepositorySymbol,
 } from '@cowprotocol/repositories';
 
+const DEFAULT_CACHE_VALUE_SECONDS = ms('2min') / 1000; // 2min cache time by default for values
+const DEFAULT_CACHE_NULL_SECONDS = ms('30min') / 1000; // 30min cache time by default for NULL values (when the repository don't know)
+
 import { Container } from 'inversify';
 import {
   SlippageService,
@@ -21,6 +24,7 @@ import {
   slippageServiceSymbol,
   usdServiceSymbol,
 } from '@cowprotocol/services';
+import ms from 'ms';
 
 function getTokenDecimals(tokenAddress: string): number | null {
   return 18; // TODO: Implement!!!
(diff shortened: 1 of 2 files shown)


apps/api/src/app/inversify.config.ts
```diff
@@ -12,6 +12,9 @@ import {
   usdRepositorySymbol,
 } from '@cowprotocol/repositories';
 
+const DEFAULT_CACHE_VALUE_SECONDS = ms('2min') / 1000; // 2min cache time by default for values
+const DEFAULT_CACHE_NULL_SECONDS = ms('30min') / 1000; // 30min cache time by default for NULL values (when the repository don't know)
```
alfetopito, line 16:
```suggestion
const DEFAULT_CACHE_NULL_SECONDS = ms('30min') / 1000; // 30min cache time by default for NULL values (when the repository isn't known)
```

apps/api/src/app/inversify.config.ts
```diff
@@ -21,6 +24,7 @@ import {
   slippageServiceSymbol,
   usdServiceSymbol,
 } from '@cowprotocol/services';
+import ms from 'ms';
```
alfetopito, line 27: Nitpick: import sorting. `ms` is imported after the const using it is defined.
Better to move the const down.

alfetopito (approved):

apps/api/src/app/inversify.config.ts
anxolin: I think we should make it automatic the sorting and cleanup of unused imports. Making an issue for this https://github.com/cowprotocol/bff/issues/76

apps/api/src/app/inversify.config.ts
anxolin: I will do in this PR to avoid merge conflicts https://github.com/cowprotocol/bff/pull/75

anxolin: 0282d7b3af53e43c1bcedea7f0fcc07821fa71f5

PR #74, opened 2024-07-30
Implement multi-chain support for cow api

anxolin:


libs/repositories/src/datasources/cowApi.ts
```diff
@@ -3,7 +3,26 @@ import createClient from 'openapi-fetch';
...
 
-export const cowApiClientMainnet = createClient<paths>({
-  baseUrl: COW_API_BASE_URL + '/mainnet',
-});
+export type CowApiClient = ReturnType<typeof createClient<paths>>;
+
+/**
+ * Name of the networks as they are in the API
+ */
+const NETWORK_NAMES: Record<SupportedChainId, string> = {
```
alfetopito, line 13:
```suggestion
const COW_API_NETWORK_NAMES: Record<SupportedChainId, string> = {
```

Then the comment is not needed IMO.

alfetopito (approved):

PR #79, opened 2024-08-02
Make logs pretty for development

anxolin:
Developing locally was a bit complicated witht he logs we have. Pino logger writes in JSON format, which makes it super hard to read:

Screenshot at Aug 02 15-57-31

Now it uses the following pretty logger for local:

Screenshot at Aug 02 15-58-05

It will check the value of `NODE_ENV` and only when it is production, it will use the default JSON format.


alfetopito (commented):
👍

PR #80, opened 2024-08-02
Add cache for notifications

anxolin:
This PR just adds a 5min cache for the notifications.

For now, the notifications don't need to be very fast. Its meant to be used for announcements. So we can afford to cache it for a few minutes. This way, we release some requests from our CMS, which has a limit in the number of requests the let us do.


anxolin:
Merging to consolidate and test the release. I address issues raised after the merge

PR #81, opened 2024-08-05
Relax env configurations

anxolin:
This PR relaxes the requirements for env configurations

It makes all envs optional providing fair default. 

For the proxies, if you don't pass the right env they won't be created. It will give you a warning though. 

It runs with no env file, it will give you just some relevant warn messages:
image


alfetopito (approved):
I can run it now, thanks!

PR #82, opened 2024-08-07
Attempt to fix issue in barn

anxolin:


alfetopito (commented):
> Don't ask me why, but this fixes the issue

But we know why.
Compressed replies need to be handled differently when consuming the stream for proper caching.
Removing this header prevents upstream (in this case Coingecko) from compressing the response.

What I don't know is how to do that 😬

PR #84, opened 2024-08-09
Implement the prediction of volatility

anxolin:
Try to predict the volatility of an order on expiration, and convert it to a slippage tolerance that would increase the order goes through.

# Proposed implementation
Understanding the tests are the best way to understand the specification of the current implementation

image

## Which info is this service using
It will get the latest 24h data points, 5min apart from Coingecko. 

It calculates the standard deviation, and with it predicts the price change at the time the order should be executed. 

## More info on the algorithm
Read the algorithm. I tried to add comments and calculate the intermediate values.

- Basically, it will calculate the slippage of each token independently
- If for some token, it doesn't know which slippage does it have (no USD estimation, no price points, etc), then we take the worst case, the slippage is the MAXIMUM
- Once we have the two slippage of the quote and base tokens, we take the worse of the two

To predict the slippage, we calculate:
- The standard deviation
- With it, we predict what is the price change in 5min (fair time for a solver to settle a trade), this will be the volatility
- The volatility is denominated in USD, so we need to normalise it so its in the relevant token

This algorithm will also acknowledge the price point are not always at regular 5min, so it will try to take into account this for making a more accurate prediction. Closer in time, higher slippage (for the same values).

For more info:
- Get the max volatility of the two tokens: https://github.com/cowprotocol/bff/blob/c7d7598ec40140f40511566a50fca64523f0735f/libs/services/src/SlippageService/SlippageServiceMain.ts#L62
- Get volatility of a token https://github.com/cowprotocol/bff/blob/c7d7598ec40140f40511566a50fca64523f0735f/libs/services/src/SlippageService/SlippageServiceMain.ts#L107

## Test
- Review the unit tests, make sure they have good coverage and expectations are sound
- Review the implementation (pointers above)
- Test endpoints for different tokens, for example
   - `USDC-DAI`: 
      - `curl -i http://localhost:3010/chains/1/markets/0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48-0x6B175474E89094C44Da98b954EedeAC495271d0F/slippageTolerance`
      - For reference I got: `50` ✅
   - `DAI-ETH`:  
      - `curl -i http://localhost:3010/chains/1/markets/0x6B175474E89094C44Da98b954EedeAC495271d0F-0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE/slippageTolerance`
      - For reference I got: `82` ✅
   - `DAI-WETH`: 
      - `curl -i http://localhost:3010/chains/1/markets/0x6B175474E89094C44Da98b954EedeAC495271d0F-0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2/slippageTolerance`
      -  For reference I got: `82` ✅

- CURL the endpoint in Arbitrum
  - `USDC-DAI`: 
    - `curl -i http://localhost:3010/chains/42161/markets/0xaf88d065e77c8cC2239327C5EDb3A432268e5831-0xDA10009cBd5D07dd0CeCc66161FC93D7c9000da1/slippageTolerance`
    - For reference I got: `50` ✅

- Execute the script to get the tolerance for a bunch of markets

```
yarn test:api:slippage
```

This will output the slippage tolerance for all those markets, and it will also persist them together with the volatility details in a JSON file. This will come handy if we want to understand how to refine the algorithm.

## Edit: 22/08/2024
The number now make more sense. I made a script to be able to test quickly various markets. The scripts print the following:
image

Additionally, I created a `test-data` directory with the prices and volatility details. My hope is that if we don't like the number of slippage we get for a given set of prices, we can take this raw data and modify the algorithm until we get the right amount we want. 

To not increase the scope further of this PR, I will review with @olgafetisova the numbers, and adjust the algorithm to our needs.


blacktieaccountant:
The issue is probably that using std. dev. is time correlated to prior periods (autoregressive) and so not a great estimation method. Probably a exponential smoothing (if possible to apply due to computational efforts) will give a better F(t+n)?

libs/services/src/SlippageService/SlippageServiceMain.ts
```diff
@@ -15,35 +23,124 @@ export class SlippageServiceMain implements SlippageService {
...
+    // Square the differences
+    const squaredDifferences = priceDifferences.map(
+      (difference) => difference ** 2
     );
 
-    return Math.ceil(bps);
+    // Calculate the variance
+    const variance =
+      squaredDifferences.reduce((acc, difference) => acc + difference, 0) /
+      prices.length;
```
alfetopito, line 123: Nitpick and tiny optimization suggestion: `variance` can be calculated in a single loop, without the intermediate `priceDifferences` and `squaredDifferences` maps.

Something like:

```suggestion
    // Calculate the variance
    const variance = prices.reduce((acc, price) => {
      // Calculate price differences between the price point and the average
      const difference = price.price - averagePrice
      // Square the difference
      const squaredDifference = difference ** 2
      // Sum squared differences
      return acc + squaredDifference
    }, 0) / prices.length
```

libs/services/src/SlippageService/SlippageServiceMain.ts
```diff
@@ -15,35 +23,124 @@ export class SlippageServiceMain implements SlippageService {
...
+      (prices[prices.length - 1].date.getTime() - prices[0].date.getTime()) /
+      (prices.length - 1);
+
+    // Points in Time for Settlement
+    const pointsForFairSettlement =
+      FAIR_TIME_TO_SETTLEMENT / averageTimeBetweenDataPoints;
+
+    // Predict variance between now and a fair settlement
+    const volatilityForFairSettlement =
+      standardDeviation * Math.sqrt(pointsForFairSettlement);
```
alfetopito, line 139: I don't get this one.
Can you explain why multiply standard dev by sqrt or points per settlement?

libs/services/src/SlippageService/SlippageServiceMain.spec.ts
```diff
@@ -1,47 +1,445 @@
...
-  getUsdPrices(
-    chainId: SupportedChainId,
-    tokenAddress: string,
-    priceStrategy: PriceStrategy
-  ): Promise<PricePoint[] | null> {
-    throw null;
-  }
-}
+const MIN_SLIPPAGE = 50;
+const MAX_SLIPPAGE = 200;
```
alfetopito, line 10: Nitpick: since it matches the implementation, what about just importing the constants from there?

libs/services/src/SlippageService/SlippageServiceMain.spec.ts
```diff
@@ -1,47 +1,445 @@
...
+
+const FIVE_MIN = ms('5min');
+const FOUR_MIN = ms('4min');
+const SIX_MIN = ms('6min');
+
+const getUsdPrice = jest.fn();
+const getUsdPrices = jest.fn();
+
+const POINTS_VOLATILITY_ZERO = getPoints([100, 100, 100, 100]);
+const POINTS_WITH_HIGH_VOLATILITY = getPoints([100, 105, 110, 120]); // 10% each 5min
```
alfetopito, line 20: First point was 5%. Adjusted to match comment.

```suggestion
const POINTS_WITH_HIGH_VOLATILITY = getPoints([100, 110, 120, 130]); // 10% each 5min
```

libs/services/src/SlippageService/SlippageServiceMain.spec.ts
```diff
@@ -1,47 +1,445 @@
...
+
+      // THEN: We get the max slippage
+      expect(result).toBe(MAX_SLIPPAGE);
+    });
+
+    it(`one of the tokens is volatile`, async () => {
+      getUsdPrices.mockImplementation(
+        async (chainId, tokenAddress, interval) => {
+          if (tokenAddress === baseTokenAddress) {
+            // GIVEN: One token is volatile
```
alfetopito, line 80:
```suggestion
            // GIVEN: One token is not volatile
```

libs/services/src/SlippageService/SlippageServiceMain.spec.ts
```diff
@@ -1,47 +1,445 @@
...
+
+      // THEN: We get the maximum slippage too
+      expect(result).toBe(MAX_SLIPPAGE);
+    });
+
+    it(`one of the tokens has no prices available`, async () => {
+      getUsdPrices.mockImplementation(
+        async (chainId, tokenAddress, interval) => {
+          if (tokenAddress === baseTokenAddress) {
+            // GIVEN: One token is volatile
```
alfetopito, line 114:
```suggestion
            // GIVEN: One token is not volatile
```

libs/services/src/SlippageService/SlippageServiceMain.spec.ts
```diff
@@ -1,47 +1,445 @@
...
+            return POINTS_VOLATILITY_ZERO;
+          } else {
+            // GIVEN: The other token is not
```
alfetopito, line 117:
```suggestion
            // GIVEN: The other token has no prices
```

libs/services/src/SlippageService/SlippageServiceMain.spec.ts
```diff
@@ -1,47 +1,445 @@
...
+        quoteTokenAddress,
+        baseTokenAddress,
+      });
+
+      // THEN: We get the maximum slippage too
+      expect(result).toBe(MAX_SLIPPAGE);
+    });
+
+    it(`if the prices change a lot`, async () => {
+      // GIVEN: The prices don't change at all
```
alfetopito, line 145: The comment is confusing and don't seem to match the test name.
The prices are volatile and are equal for both tokens.

libs/services/src/SlippageService/SlippageServiceMain.spec.ts
```diff
@@ -1,47 +1,445 @@
...
+        quoteTokenAddress,
+      });
+
+      // THEN: We get the maximum slippage
+      expect(result).toBe(MAX_SLIPPAGE);
+    });
+
+    it(`if the asset is volatile`, async () => {
+      // GIVEN: The prices have high volatility
+      getUsdPrices.mockResolvedValue(getPoints([100, 100.02, 100.04, 100.06]));
```
alfetopito, line 161: 0.02% is high?

libs/services/src/SlippageService/SlippageServiceMain.spec.ts
```diff
@@ -1,47 +1,445 @@
...
+      });
+
+      // THEN: We get the maximum slippage
+      //    AVG = (100 + 100.02 + 100.04 + 100.06)/4 = 100.03
+      //    VARIANCE = ((100 - 100.03)**2 + (100.02 - 100.03)**2 + (100.04 - 100.03)**2 + (100.06 - 100.03)**2) / 4 = 0.0005
+      //    STDDEV = sqrt(0.0005) = 0.02236067977
+      //    Volatility Fair Settlement (USD) = 0.02236067977 * sqrt(1) = 0.02236067977
+      //    Volatility Fair Settlement (Token) = 0.02236067977 / 1 = 0.02236067977
+      //    Slippage BPS = ceil(0.02236067977 * 10000) = 224
+      //    Adjusted Slippage = MAX = 200
```
alfetopito, line 180: Thanks, this helps visualize it!

I still don't get line 177, same as I commented in the code.

libs/services/src/SlippageService/SlippageServiceMain.spec.ts
```diff
@@ -1,47 +1,445 @@
...
+      });
+
+      // THEN: We get the the calculated slippage
+      //    AVG = (100 + 100.01 + 100.02 + 100.03)/4 = 100.015
+      //    VARIANCE = ((100 - 100.015)**2 + (100.01 - 100.015)**2 + (100.02 - 100.015)**2 + (100.03 - 100.015)**2) / 4 = 0.000125
+      //    STDDEV = sqrt(0.000125) = 0.01118033989
+      //    Volatility Fair Settlement (USD) = 0.01118033989 * sqrt(1) = 0.01118033989
+      //    Volatility Fair Settlement (Token) = 0.01118033989 / 0.9 = 0.01242259988
+      //    Slippage BPS = ceil(0.01242259988 * 10000) = 125
+      //    Adjusted Slippage = 102
```
alfetopito, line 289:
```suggestion
      //    Adjusted Slippage = 125
```

libs/services/src/SlippageService/SlippageServiceMain.spec.ts
```diff
@@ -1,47 +1,445 @@
...
+
+      // WHEN: Get slippage (inverting the tokens)
+      const resultInverted = await slippageService.getSlippageBps({
+        chainId,
+        baseTokenAddress,
+        quoteTokenAddress,
+      });
+
+      // THEN: The result should be the same (worst of the two)
+      expect(resultInverted).toBe(200);
```
alfetopito, line 430: Nitpick: several places `200` is used when indicating the max slippage, but there's a const for that.

alfetopito (commented):
Reviewed the code, did not test it yet

alfetopito:
Found the reason for the difference between WETH and ETH prices

image

I added some local debug logs. Would you like me to push this?
image
image

apps/api/src/app/routes/__chainId/markets/__baseTokenAddress-__quoteTokenAddress/slippageTolerance.ts
```diff
@@ -65,20 +77,24 @@ const root: FastifyPluginAsync = async (fastify): Promise<void> => {
...
+        JSON.stringify(request.query)
       );
-      const slippageBps = await slippageService.getSlippageBps(
+      const slippageBps = await slippageService.getSlippageBps({
+        chainId,
         baseTokenAddress,
-        quoteTokenAddress
-      );
+        quoteTokenAddress,
+      });
```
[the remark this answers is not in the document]
anxolin: Yep, not for now. I don't even know if we will want them for sure. I was experimenting

libs/services/src/SlippageService/SlippageServiceMain.spec.ts
anxolin: yes, its 0.02% every 5min

According to the calculation, we get `224 BPS`  which I think is volatile

libs/services/src/SlippageService/SlippageServiceMain.ts
anxolin: I tried to commit this, but said the code is outdated and wouldn't let me. I just manually added your changes

libs/services/src/SlippageService/SlippageServiceMain.ts
anxolin: This is the trick on how to use the variance to predict the volatility at any point in time

Here I calculate how many point points are needed between now and a fair time for settlement, and then I need to multiply the variance by sqrt of the number of points

If I want to predict whats the volatility in 15min, I need to multiply by `sqrt(3)` cause each point is ~5min --> so multiply by `1.7320508076`

libs/services/src/SlippageService/SlippageServiceMain.spec.ts
anxolin: I hope is more clear now. I tried to explain in the thread above, and added more comments

apps/api/scripts/test-slippage.js
```diff
@@ -0,0 +1,139 @@
+const fs = require('fs');
```
anxolin, line 1: this script should make it easier to test with real data we get the right slippage tolerance, and we can reproduce easily the calculations

I will use the output files to refine the algorithms defining a test that would ensure we get the right value for a realistic example of 288 prices

apps/api/src/app/inversify.config.ts
```diff
@@ -107,7 +107,7 @@ function getApiContainer(): Container {
   // Services
   apiContainer
     .bind<SlippageService>(slippageServiceSymbol)
-    .to(SlippageServiceMock);
+    .to(SlippageServiceMain);
```
anxolin, line 110: Now we use the new implementation!

apps/api/src/app/routes/__chainId/markets/__baseTokenAddress-__quoteTokenAddress/slippageTolerance.ts
```diff
@@ -31,6 +35,18 @@ const routeSchema = {
...
 
+const queryStringSchema = {
+  type: 'object',
+  properties: {
+    orderKind: { type: 'string', enum: ['buy', 'sell'] },
+    partiallyFillable: { type: 'boolean' },
+    sellAmount: { type: 'string' },
+    buyAmount: { type: 'string' },
+    expirationTimeInSeconds: { type: 'number' },
+    feeAmount: { type: 'string' },
```
anxolin, line 46: For now, let's ignore these params

apps/api/src/app/routes/__chainId/markets/__baseTokenAddress-__quoteTokenAddress/slippageTolerance.ts
```diff
@@ -72,18 +89,63 @@ const root: FastifyPluginAsync = async (fastify): Promise<void> => {
...
+  );
+
+  fastify.get<{
+    Params: RouteSchema;
+    Reply: {
+      baseToken: VolatilityDetails | null;
+      quoteToken: VolatilityDetails | null;
+    };
+  }>(
+    '/volatilityDetails',
```
alfetopito, line 117: Ah thought that just crossed my mind, feel free to ignore.

I'm not very happy with an endpoint just for debugging.
What about having an optional (maybe hidden) parameter in the regular call that returns all of this data with the slippage request?

libs/services/src/SlippageService/test-data/1-COW-WETH__slippageTolerance.json
alfetopito, line 1: Were these files added to the PR intentionally? If so, should it really?

anxolin: Next PR deletes them and add new ones. I should have not add them. But let me leave them, so I don't get merge conflicts

apps/api/src/app/routes/__chainId/markets/__baseTokenAddress-__quoteTokenAddress/slippageTolerance.ts
anxolin: Yes, is a fair point, but also I don't love adding this additional data in the slippage tolerance endpoint, cause this is the one people will use and I don't want to expose these internal details. 

Although I used the volatilityDetails for a market for debugging, I think it doesn't harm to have it there. It returns nice information about a market volatility. 

I can also hide it in swagger to make it "not official" would that help?

anxolin: Leandro, I will merge and address things as follow ups, cause I want to test this in CoW Swap so I want to deploy it to staging. 

This endpoint is not used in CoW Swap, so I hope the resolution of this one won't affect there

alfetopito (commented):
👍

PR #87, opened 2024-10-04
Add Simulate Bundle and find top token holders endpoints

yvesfracari:
Context: https://github.com/cowprotocol/cowswap/pull/4943

Note: Couldn't make the body on the tenderly endpoint render on swagger, but it still works 😅


yvesfracari:
Closing due to #88

PR #90, opened 2024-10-18
fix: Simulate bundle error message

yvesfracari:


apps/api/src/app/routes/__chainId/simulation/simulateBundle.ts
```diff
@@ -123,7 +123,7 @@ const root: FastifyPluginAsync = async (fastify): Promise<void> => {
         );
 
       if (simulationResult === null) {
-        reply.code(404).send({ message: 'Token holders not found' });
+        reply.code(404).send({ message: 'Build simulation error' });
```
alfetopito, line 126: In this case - it's an error an not something that was not found - shouldn't the return code be something other than `404`? Maybe a generic `400` if this is due to user input or even a `500` if this is something on the server side.
I don't have the full context of this in mind, so not sure what's more appropriate.

yvesfracari: You are definitely right. I think that 400 is more appropriate.

PR #93, opened 2024-10-23
feat: add `gasUsed` data on tenderly simulation endpoint

yvesfracari:
Run the automatic tests with:

`nx test repositories --testFile=libs/repositories/src/SimulationRepository/SimulationrepositoryTenderly.test.ts`

By the way, as you can see the gas parameter is returned even in failed simulations, so maybe we don't need to make the hook app pass a `maxGasLimit` parameter.


alfetopito:
> maybe we don't need to make the hook app pass a maxGasLimit parameter.

I think we still do because in case of a revert along the way, the failed simulation might use just a fraction of the gas needed

alfetopito (approved):

PR #94, opened 2024-10-29
Feat add token holder from moralis

yvesfracari:
To test:

- Set the new env variable by creating one Moralis API Key.
- Run the new tests with: `nx test repositories --testFile=libs/repositories/src/TokenHolderRepository/TokenHolderRepositoryMoralis.spec.ts`


alfetopito (approved):
Working great!!

PR #99, opened 2024-11-25
feat(yield): return liquidity pools info

shoom3301:
There are two new endpoints:

https://bff.barn.cow.fi/42161/yield/pools-average-apr - returns average APR per provider
https://bff.barn.cow.fi/42161/yield/pools - returns pools info (tvl, apr, etc.)

BFF endpoint instead of [this mock](https://github.com/cowprotocol/cowswap/blob/562d0207d1acf4e1735c4b3f629ff63dd65d3725/apps/cowswap-frontend/src/modules/yield/updaters/PoolsInfoUpdater/mockPoolInfo.ts).
This data is used for Vampire attack.

image


apps/api/src/app/routes/__chainId/yield/getPoolsAverageApr.ts
```diff
@@ -0,0 +1,61 @@
...
+
+        acc[projectName] = +val.average_apr.toFixed(6)
+
+        return acc
+      }, {})
+
+      reply.status(200).send(averageApr);
+    }
+  );
+};
```
anxolin, line 59: Are you applying any caching? 

I think if you add: `reply.header(CACHE_CONTROL_HEADER, getCacheControlHeaderValue(600))` you will be able to cache in redis the request

apps/api/src/app/routes/__chainId/yield/getPoolsInfo.ts
```diff
@@ -0,0 +1,58 @@
...
+type ErrorSchema = FromSchema<typeof errorSchema>;
+type BodySchema = FromSchema<typeof poolsInfoBodySchema>;
+
+const root: FastifyPluginAsync = async (fastify): Promise<void> => {
+  fastify.post<{
+    Params: RouteSchema;
+    Reply: SuccessSchema | ErrorSchema;
+    Body: BodySchema;
+  }>(
+    '/getPoolsInfo',
```
anxolin, line 20: This method and path is not following very match the restful structure
i.e. https://restfulapi.net/resource-naming

what about `GET /pools/OxB7ECB2AA52AA64A717180E030241BC75CD946726`

anxolin (commented):
nice one! 

Just a couple of nits

apps/api/src/app/routes/__chainId/yield/getPoolsInfo.ts
shoom3301: Agree, I just inhertted this style from the existing paths:
image

Anyway, I will rename it

shoom3301: Renamed

image

apps/api/src/app/routes/__chainId/yield/getPoolsAverageApr.ts
shoom3301: Thank you! Added

anxolin:
I stil don't get what's the API naming convention here. Maybe we can discuss it synchronously while we deploy this

anxolin (commented):
Other than the endpoint names, where to be honest I don't care too much anymore I have some small concern :) 

One concern with this PR is that, so far we made that BFF can be easily cloned and launched locally. 

This PR introduces one dependency. We need to provide the CoW AMM DB details, or it will try to connect to `127.0.0.1:5432`

image

For example, for coingecko proxy we made the env optional, and we just printed an error if its missconfigured 

https://github.com/cowprotocol/bff/blob/feat/com-amm-competitor-api/apps/api/src/app/routes/proxies/coingecko/index.ts#L24

Would it make sense to do something like this here? 
We can throw an error if the env is not set up for example and show in the logs what's the issue

I will give you the soft approve as I think this PR is quite good.

tsconfig.base.json
```diff
@@ -20,6 +20,7 @@
     "strict": true,
     "strictNullChecks": true,
     "alwaysStrict": true,
+    "strictPropertyInitialization": false,
```
anxolin, line 23: Why do we want this? https://www.typescriptlang.org/tsconfig/#strictPropertyInitialization

shoom3301: image

There is a TS error. But I realized that it works in `twap` app, so I just copy-pasted tsconfig from here.

shoom3301:
image

@anxolin I've enhanced the db connection error handling

alfetopito (approved):

anxolin:
> @anxolin I've enhanced the db connection error handling

Thanks a lot, looks way better now. Sorry it took me a while to re-review this

PR #101, opened 2024-12-17
chore: add pg to api deps

shoom3301:
`typeorm` library doesn't include database provider dependencies, so you need to install them by yourself.
We have `pg` dependency in the root `package.json` and it works well when you build and run the app locally, it even works if you do `docker compose`!

Anyway, in aws we get this error:
```
{"type":"DriverPackageNotInstalledError","message":"Postgres package has not been found installed. Try to install it: npm install pg --save","stack":"DriverPackageNotInstalledError: Postgres package has not been found installed. Try to install it: npm install pg
```

@ahhda found a key of the problem, in the container we have a package.json of `/app/api` which is generated by `nx` while building and it doesn't include `pg` dependency.
I assume, `nx` only collects dependencies that explicitly are in use in the app.
To fix that I've initialized `package.json` in `api` app and explicitly added `pg` dependency there.

Before:

image

After:

image


ahhda (approved):
Works well 👍 

Would be great to also solve the root cause so that we don't have the same issue with other BFF images. Adding dependency to root `package.json` should be reflected in the sub projects `package.json` upon build.

shoom3301:
>Would be great to also solve the root cause so that we don't have the same issue with other BFF images

@ahhda I think my assumption from the PR description is right, this is just a nuance how `typeorm` works with db providers and how `nx` collects dependencies for apps.
Anyway, I will add this information to README.

apps/api/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "api",
+  "version": "1.0.0",
+  "description": "",
+  "main": "./main.js",
+  "scripts": {},
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "pg": "^8.13.1"
```
alfetopito, line 10: My only worry on this is to have another package json to keep track of

alfetopito (commented):
👍

PR #103, opened 2025-01-13
fix: change tenderly transaction type to optional on tenderly simulation

yvesfracari:
This PR only changes the type to optional and returns the null gas used since the value will not be presented

### DIFF
--- a/libs/repositories/src/SimulationRepository/SimulationRepository.ts
+++ b/libs/repositories/src/SimulationRepository/SimulationRepository.ts
@@ -16,7 +16,7 @@ export interface SimulationData {
   // { [address: string]: { [token: string]: balanceDiff: string } }
   // example: { '0x123': { '0x456': '100', '0xabc': '-100' } }
   cumulativeBalancesDiff: Record<string, Record<string, string>>;
-  gasUsed: string;
+  gasUsed?: string;
 }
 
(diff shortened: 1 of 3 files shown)


libs/repositories/src/SimulationRepository/SimulationRepository.ts
```diff
@@ -16,7 +16,7 @@ export interface SimulationData {
   // { [address: string]: { [token: string]: balanceDiff: string } }
   // example: { '0x123': { '0x456': '100', '0xabc': '-100' } }
   cumulativeBalancesDiff: Record<string, Record<string, string>>;
-  gasUsed: string;
+  gasUsed?: string;
```
alfetopito, line 19: Is this needed? Seems unrelated since the issue is when `transaction` is null

alfetopito (approved):
One small comment, otherwise looks good

libs/repositories/src/SimulationRepository/SimulationRepository.ts
yvesfracari: This is needed because of the change on the line 65 of `libs/repositories/src/SimulationRepository/SimulationRepositoryTenderly.ts` file

anxolin (approved):

PR #105, opened 2025-01-20
debug: increase logs of simulateBundle endpoint

yvesfracari:


apps/api/src/app/routes/__chainId/simulation/simulateBundle.ts
```diff
@@ -130,23 +130,32 @@ const root: FastifyPluginAsync = async (fastify): Promise<void> => {
...
-        `Post Tenderly bundle of ${request.body.length} simulation on chain ${chainId}`
-      );
+        const simulationResult =
+          await tenderlyService.postTenderlyBundleSimulation(
+            chainId,
+            request.body
+          );
+
+        if (simulationResult === null) {
+          reply.code(400).send({ message: 'Build simulation error' });
```
anxolin, line 147: This was here before, but what does `null` mean, and why is a `400` is an issue from the user?

anxolin (commented):
I would also add some error handling in the repository. Handle and retrow logic.

At minimum we can log the payload and relevant response HTTP parameters
https://github.com/cowprotocol/bff/blob/main/libs/repositories/src/SimulationRepository/SimulationRepositoryTenderly.ts#L36

Alternativelly, and probably better, is to model some Error (ApiCallError) that incorporates all the relevant data we want to handle, and then the controllers handles and logs this error properly. This approach is nicer cause you don't marry to the specific logging methon, and then other similar repositories can do the same

apps/api/src/app/routes/__chainId/simulation/simulateBundle.ts
yvesfracari: It means that the Tenderly couldn't simulate the transaction. This happens often on the "build your own hook" where the user manually encodes the tx data.

We used a general 400 error but maybe we could change it to 422.

anxolin (approved):
Thanks!

apps/api/src/app/inversify.config.ts
```diff
@@ -47,6 +42,20 @@ import {
...
+
+const CACHE_TOKEN_INFO_SECONDS = ms('24h') / 1000; // 24h
+
+// Configure the logger
+const logger = pino({
+  level: process.env.LOG_LEVEL || 'info',
+  formatters: {
+    level: (label) => ({ level: label }),
+  },
+});
```
anxolin, line 58: @yvesfracari why do we add the log config here, shouldn't we use the same one as https://github.com/cowprotocol/bff/blob/main/apps/api/src/main.ts#L5 ?

anxolin: I address it, cause I want to deploy a new release of bff with some fixes for the logging and I want to fix this small detail. Let me know if it makes sense

PR #115, opened 2025-04-15
fix: do not assume 18 decimals when it can't be fetched

alfetopito:
# Summary

Do not assume token has 18 decimals when the values can't be fetched from the contract.
It can have unwanted consequences, like a bad token estimation.

When not found, return a 400.

# Testing

- Added unit tests.

### DIFF
--- a/libs/repositories/src/UsdRepository/UsdRepositoryCow.ts
+++ b/libs/repositories/src/UsdRepository/UsdRepositoryCow.ts
@@ -27,8 +27,10 @@ export class UsdRepositoryCow extends UsdRepositoryNoop {
     }
     const erc20 = await this.erc20Repository.get(chainId, tokenAddress);
     const tokenDecimals = erc20?.decimals;
+
     if (tokenDecimals === undefined) {
-      throw new Error('Token decimals not found for ' + tokenAddress);
+      console.info('Token decimals not found for ' + tokenAddress);
+      return null;
     }
 
     // Get native price for USDC (in ETH/xDAI)
--- a/libs/repositories/src/UsdRepository/UsdRepositoryCow.ts
+++ b/libs/repositories/src/UsdRepository/UsdRepositoryCow.ts
@@ -1,11 +1,11 @@
-import { injectable } from 'inversify';
-import { UsdRepositoryNoop } from './UsdRepository';
-import { OneBigNumber, TenBigNumber, USDC, ZeroBigNumber } from '../const';
 import { SupportedChainId } from '@cowprotocol/shared';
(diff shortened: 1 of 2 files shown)


libs/repositories/src/UsdRepository/UsdRepositoryCow.ts
```diff
@@ -26,9 +26,11 @@ export class UsdRepositoryCow extends UsdRepositoryNoop {
       return null;
     }
     const erc20 = await this.erc20Repository.get(chainId, tokenAddress);
-    const tokenDecimals = erc20?.decimals || 18;
-    if (tokenDecimals === null) {
-      throw new Error('Token decimals not found for ' + tokenAddress);
+    const tokenDecimals = erc20?.decimals;
+
+    if (tokenDecimals === undefined) {
+      console.info('Token decimals not found for ' + tokenAddress);
```
alfetopito, line 32: Is there a better logger maybe?

libs/repositories/src/UsdRepository/UsdRepositoryCow.ts
```diff
@@ -26,9 +26,11 @@ export class UsdRepositoryCow extends UsdRepositoryNoop {
...
+      return null;
```
alfetopito, line 33: Return null instead of throwing to return 400 instead of 500.

libs/repositories/src/UsdRepository/UsdRepositoryCow.ts
anxolin: We do have better logger. https://github.com/cowprotocol/bff/blob/hackathon-mooo/apps/api/src/logger.ts#L26 

Is this something we really want to log?

anxolin (approved):

libs/repositories/src/UsdRepository/UsdRepositoryCow.ts
alfetopito: We are blind when it comes to why a token has no price otherwise.
In fact, right now I can't tell why on prod it returns `315438595.8198187` while on local it returns `0.031570713153867405`.
Yes, the assumed decimals to be 18 play a role, but why it loads no decimals in the first place is the unknown part.

And thanks, I'll use that logger :)

anxolin: I think the key is here 
image

Thanks for adding the logger

alfetopito: Yes. The question that I have is, why did it fail only for decimals?

PR #116, opened 2025-04-16
fix: handle errors

anxolin:
This PR handles the error that's making crashlooping in prod. We just crash if the queue message has not the right format

Also, prints some useful info on which part throws an error

### DIFF
--- a/apps/telegram/src/main.ts
+++ b/apps/telegram/src/main.ts
@@ -53,15 +53,41 @@ async function getSubscriptions(
   return SUBSCRIPTION_CACHE.get(account) || [];
 }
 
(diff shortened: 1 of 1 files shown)


apps/telegram/src/main.ts
```diff
@@ -104,7 +130,9 @@ async function connect() {
     NOTIFICATIONS_QUEUE,
     async (msg) => {
       if (msg !== null) {
-        onNewMessage(channel, msg);
+        onNewMessage(channel, msg).catch((error) =>
+          console.error('Error processing queue message', error)
+        );
```
anxolin, line 135: this is the key part. Not throw if message is wrong

alfetopito (approved):

PR #119, opened 2025-04-17
feat: refactor producers

anxolin:
This PR refactor our producer code so we can easily add new producers easily

- Creates a basic repository to make it easy to push messages to a queue
- Create a util to make it easier to run forever and call periodically the fetch/push function
- Creates a CMS notification producer which receives the repository and simplifies its logic
- Re-arrange the main producer entry point so we can accommodate new independent producers

Also, it changes the message in the queue from accepting a single notification, to accepting an array of them.
This would help with the transactionality, although the current implementation is still not great as we can get an error halfway of sending messages. So it would be ideal to have re-attempt logics.


alfetopito (approved):
Just a few nitpicks

PR #121, opened 2025-04-17
feat: create db and update

anxolin:
Creates table in the DB (i include the SQL to create them)

image

I included in the trade notification producer the part of:
- getting the state from the db
- getting the view client for the current network
- query the last block
- updating the db with the latest block

## Summary by CodeRabbit

- **New Features**
  - Introduced persistent tracking of notification indexer state for improved reliability and scalability.
  - Added support for handling multiple blockchain networks by tracking state per chain.

- **Bug Fixes**
  - Ensured the database service is started as part of the setup instructions.

- **Documentation**
  - Updated setup instructions to include starting the database container.

- **Chores**
  - Upgraded PostgreSQL package and added type definitions for improved development experience.


apps/notification-producer/db.sql
```diff
@@ -0,0 +1,25 @@
+CREATE TABLE notifications_indexer_state (
+  key TEXT NOT NULL,
+  chainId INTEGER,
```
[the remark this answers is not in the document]
anxolin: it can be null if the indexing applies to any network. For example, governance notifications

libs/cms-api/src/index.ts
```diff
@@ -214,8 +214,7 @@ async function getTelegramSubscriptionsForAccounts({
 
   if (error) {
     console.error(
-      `Error ${response.status} getting telegram subscriptions: ${response.url}. Page${page}`,
-      error
+      `Error ${response.status} getting telegram subscriptions: ${response.url}. Page${page}`
```
alfetopito, line 217: Sure we don't want to log the error with the stack trace?

alfetopito (commented):
One comment to add to the ones from coderabbit.

libs/cms-api/src/index.ts
anxolin: yes, cause we retrow

PR #123, opened 2025-04-18
feat: refactor IoC and format amounts

anxolin:
This PR tries to format the amounts of the trade events to print them nicely and show the symbol.

image

In order to do it, I moved the IoC to a shared lib. For now services, but probably makes sense to move it either to some other lib, or let each lib export their factories (probably last makes more sense).

Anyways, the producers uses the ERC20repo now to implement the formatting


apps/api/src/main.ts
```diff
@@ -1,6 +1,6 @@
 import Fastify from 'fastify';
 import { app } from './app/app';
-import { logger } from './logger';
+import { logger } from '../../../libs/shared/src/logger';
```
shoom3301, line 3: Nitpick: Would be great to add TS alias in the future

anxolin: I will do it now, cause I will forget

shoom3301, line 3:
```suggestion
import { logger } from '@cowprotocol/shared';
```

shoom3301 (approved):

alfetopito:
Isn't this a problem?
image

PR #126, opened 2025-04-22
feat: create datasources and repository for push

anxolin:
This PR continues with the cleanup after the hackathon (follow up to https://github.com/cowprotocol/bff/pull/125)

## API error
Tries to solve this error in the API. 

Technically the API doesn't need the queue, but is requiring the env. This is because it was asserting the presence of the env just by importing the `factory.ts` file.

This PR delays the check to the actual construction of the queue connection. 

image

## Improves the PushNotificationRepository

It moves there most of the logic that was done before in the 2 executable files: `telegram` and `producers`, simplifying their logic.

Now the repository handles also subscriptions, and posting messages to the queue, hiding the re-attempt logics and confirmation/rejection of the consumption of the message.

Now the push notification repository has these methods, which makes it very easy to work with:
```ts
export interface PushNotificationsRepository {
  connection: ConnectToChannelResponse | null;

  connect(): Promise<ConnectToChannelResponse>;
  close(): Promise<void>;

  send(notifications: PushNotification[]): Promise<void>;
  subscribe(
    callback: (notification: PushNotification) => Promise<boolean>
  ): Promise<QueueSubscription>;
  ping(): Promise<boolean>;
}
```

## Moves `telegram` and `rabbitMq` connection to `datasources`
Moves these instantiation and check of the env vars to the `datasources`, this way, is easier for repositories to get the access to the data and reuse datasources:

image

## Applies also some misc improvements
- Use UUID for tracking messages
- Improve type assertion for messages (check not only the presence of properties, but also the type)
- Improve and simplify the parsing of messages

## Summary by CodeRabbit

- **New Features**
  - Introduced a push-based notification subscription model, allowing real-time delivery of notifications.
  - Added robust connection management and retry logic for notification delivery.
  - Provided utilities for runtime validation and serialization of notification data.
  - Added support for initializing and managing a singleton Telegram bot instance.

- **Refactor**
  - Simplified notification sending and subscription interfaces.
  - Transitioned from manual RabbitMQ queue handling to a streamlined repository-based approach.
  - Streamlined logging and error handling for notification processing.

- **Chores**
  - Updated exports to include new RabbitMQ and Telegram data source utilities.

### DIFF
--- a/apps/notification-producer/src/producers/cms/CmsNotificationProducer.ts
+++ b/apps/notification-producer/src/producers/cms/CmsNotificationProducer.ts
@@ -88,7 +88,7 @@ export class CmsNotificationProducer implements Runnable {
     await this.props.pushNotificationsRepository.connect();
 
     // Post notifications to queue
-    this.props.pushNotificationsRepository.sendNotifications(pushNotifications);
+    this.props.pushNotificationsRepository.send(pushNotifications);
     this.pendingNotifications.clear();
   }
 }
--- a/apps/notification-producer/src/producers/trade/TradeNotificationProducer.ts
+++ b/apps/notification-producer/src/producers/trade/TradeNotificationProducer.ts
@@ -122,7 +122,7 @@ export class TradeNotificationProducer implements Runnable {
     );
 
     // Post notifications to queue
-    this.props.pushNotificationsRepository.sendNotifications(notifications);
+    this.props.pushNotificationsRepository.send(notifications);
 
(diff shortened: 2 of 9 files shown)


anxolin:
@alfetopito @shoom3301 I tested this locally and it works. I will merge to check if this fixes the issue in barn. I might need to do a few iterartions cause there might be more errors, and I want to iterate fast, but feel free to add your comments

PR #128, opened 2025-04-23
Apply automatic migrations

anxolin:
Automatically apply the repository migrations.

## background on this PR
I was not planning to do this PR, as I just need to create a single table, and I was fine with applying the migrations to a remote db, but I don't have write permissions to production or staging

There might be many repositories using the DB, and can also be used in different apps. If I try to apply migrations in all apps boot, there could be a race condition. My first approach was using docker and apply the migration when we deploy the services, but later I realised there was a simple way by using the API. 

The API is deployed each time there's a new BFF version, so now in its boot will apply any un-applied migration. 
It was also a bit of not a straight line to make it work. I will try to merge this PR and test this in AWS to see if it works well there. 

# Changes

## Namespaces for different ORM
The API had an orm of `typeorm` 

## Automatically apply migrations of the repositories 
The API defines a new plugin that applies the migrations. 

image

## Summary by CodeRabbit

- **New Features**
  - Introduced a new plugin to manage PostgreSQL database connections and migrations, with improved logging and error handling.
- **Refactor**
  - Updated database connection logic for improved maintainability and clarity.
  - Repository access for analytics endpoints now uses a dedicated analytics namespace.
- **Chores**
  - Enhanced code formatting for readability across several modules.

### DIFF
--- a/apps/api/src/app/plugins/orm-analytics.ts
+++ b/apps/api/src/app/plugins/orm-analytics.ts
@@ -11,25 +11,30 @@ export default fp(async function (fastify: FastifyInstance) {
     database: fastify.config.COW_ANALYTICS_DATABASE_NAME,
     username: fastify.config.COW_ANALYTICS_DATABASE_USERNAME,
     password: fastify.config.COW_ANALYTICS_DATABASE_PASSWORD,
-  }
+  };
 
-  const dbParamsAreInvalid = Object.values(dbParams).some((v) => Number.isNaN(v) || v === undefined);
+  const dbParamsAreInvalid = Object.values(dbParams).some(
+    (v) => Number.isNaN(v) || v === undefined
+  );
 
   if (dbParamsAreInvalid) {
-    console.error('Invalid CoW Analytics database parameters, please check COW_ANALYTICS_* env vars');
-    return
+    console.error(
+      'Invalid CoW Analytics database parameters, please check COW_ANALYTICS_* env vars'
+    );
+    return;
   }
 
   fastify.register(typeORMPlugin, {
     ...dbParams,
+    namespace: 'analytics',
     type: 'postgres',
     entities: [PoolInfo],
     ssl: true,
     extra: {
       ssl: {
-        rejectUnauthorized: false
-      }
-    }
+        rejectUnauthorized: false,
+      },
+    },
   });
 
   fastify.ready((err) => {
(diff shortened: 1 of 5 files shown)


apps/api/src/app/plugins/orm-analytics.ts
```diff
@@ -11,32 +11,37 @@ export default fp(async function (fastify: FastifyInstance) {
...
-    return
+    console.error(
+      'Invalid CoW Analytics database parameters, please check COW_ANALYTICS_* env vars'
+    );
+    return;
   }
 
   fastify.register(typeORMPlugin, {
     ...dbParams,
+    namespace: 'analytics',
```
anxolin, line 29: namespacing helps to differenciate between the different `typeorm` instances. This one is for the analytics.

apps/api/src/app/plugins/orm-repositories.ts
```diff
@@ -0,0 +1,67 @@
...
+import { getDatabaseParams } from '@cowprotocol/repositories';
+import { readdir } from 'fs/promises';
+
+export default fp(async function (fastify: FastifyInstance) {
+  const isProduction = process.env.NODE_ENV === 'production';
+
+  const migrationsDir = resolve(
+    __dirname,
+    '../../../../../libs/repositories/src/migrations'
+  );
```
anxolin, line 17: It took me a bit to arrive here, when we run the API using nx, it depends on the build. This PR adds a build step for the repositories too.  This way, we can find the migrations and use them.

apps/api/src/app/plugins/orm-repositories.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  fastify.register(typeORMPlugin, {
+    ...getDatabaseParams(),
+    namespace: 'repositories',
+    type: 'postgres',
+    migrations: [`${migrationsDir}/*.js`],
+    migrationsTableName: 'migrations_repositories',
+    entities: [],
+    ssl: isProduction,
+    extra: isProduction
```
anxolin, line 27: use ssl only in prod

libs/repositories/src/datasources/orm/postgresOrm.ts
```diff
@@ -2,25 +2,28 @@
 import { DataSource } from 'typeorm';
 
 import assert from 'assert';
-import { IndexerState } from '../../database/IndexerState.entity';
 
-export function createNewPostgresOrm(): DataSource {
+export function getDatabaseParams() {
```
anxolin, line 6: just convenient to share the credentials to be used in the `typeorm` API config

anxolin:
@coderabbitai review

PR #140, opened 2025-06-05
feat: remove dependency with db

anxolin:


alfetopito (approved):
Approved, but see my comment about CMS

PR #145, opened 2025-07-23
feat: add dune support and export hooks

anxolin:
This PR adds a repository to access dune queries, or execute them and get results.

It also exposes hook!

## Motivation
We don't have a way to easily read in CoW Swap from Dune. 

Soon we will need to implement the affiliate program, so I did this experimental PR to see if I can get the hooks (since we worked recently in creating some query to return the hooks for the bridge and swap feature 😃

## Scope
This is experimental, I don't plan to use it yet in CoW Swap, but leave this easy to add any data we want in CoW Swap now, I think it will be very handy. 

## Repository 
It can:
- `getQueryResults`: Return the cached last execution for a given query. Allows pagination
- `executeQuery`: Execute a query. The query returns an `execution_id`
- `getExecutionResults`: Return the result for a given `execution_id` if ready
- `waitForExecution`: Waits for a `execution_id` to be ready, and then return the result

## Hooks Service

- `getHooks`: Returns hooks given some parameters (`blockchain` and `period`). 
- `getLatestHooks`: Returns all the cached hooks for the last 30 days. This is kind of experimental query. In this PR I'm only interested on implementing a generic way to access dune and extract data so it can be exported nicely in our endpoints. 

### /hooks
Screenshot at Jul 23 21-43-41

### /hooks/latest
image

## Caveats
The parametrization of `getHooks` is not working for me. I pass the parameters to `executeQuery` as I understood from the API, but somehow they are ignored. It's likely something silly, and since this is easily fixable by having a query with the parameters we want, and we don't plan to use this right now I didn't loose more time on this. 

## Test

add API key for dune
```
DUNE_API_KEY=
```

Start service:
```
yarn start
```

Go to:
http://localhost:3001

Get all hooks:

http://localhost:3001/hooks/latest?limit=1000&offset=0

Screenshot at Jul 23 22-07-54

## Summary by CodeRabbit

* **New Features**
  * Optional, feature-flagged Dune integration and a GET /hooks endpoint to fetch hooks by blockchain/period with pagination, validation, and 5-minute caching.
  * New service and repository implementations plus public service/repository exports for use by the API.

* **Tests**
  * Extensive unit tests for the Dune client (requests, polling, timeouts, validation) and the Hooks service flows.

* **Chores**
  * Added DUNE_API_KEY placeholder to env example and a factory to construct the Dune client.

* **Documentation**
  * Added contributor guidance (local route/schema conventions and pre-handoff checks).

### DIFF
--- a/apps/api/src/app/routes/hooks.ts
+++ b/apps/api/src/app/routes/hooks.ts
@@ -55,11 +55,6 @@ const hooks: FastifyPluginAsync = async (fastify): Promise<void> => {
       },
     },
     async function (request, reply) {
-      reply.header(
-        CACHE_CONTROL_HEADER,
-        getCacheControlHeaderValue(CACHE_SECONDS)
-      );
-
       try {
         const hooksService = apiContainer.get<HooksService>(hooksServiceSymbol);
         const hooks = await hooksService.getHooks({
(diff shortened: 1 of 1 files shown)


kernelwhisperer:
> The parametrization of getHooks is not working for me. I pass the parameters to executeQuery as I understood from the API, but somehow they are ignored. It's likely something silly, and since this is easily fixable by having a query with the parameters we want, and we don't plan to use this right now I didn't loose more time on this.

The fix was to send the params in the request body

PR #153, opened 2025-09-02
feat: add moralis balances api support

limitofzero:
## Summary by CodeRabbit

* **New Features**
  * Added GET /:chainId/address/:address/balances endpoint — input validation, short caching (~5s), 404 when no balances, and clear 5XX responses.
  * Exposed a Token Balances Service for programmatic access.
  * Integrated a Moralis-backed balances provider (with a fallback/no-op implementation) and a factory to obtain it.

### DIFF
--- a/libs/services/src/factories.ts
+++ b/libs/services/src/factories.ts
@@ -12,6 +12,8 @@ import {
   Erc20RepositoryFallback,
   Erc20RepositoryNative,
   Erc20RepositoryViem,
+  ExpiredOrdersRepository,
+  ExpiredOrdersRepositoryPostgres,
   getViemClients,
   IndexerStateRepository,
   IndexerStateRepositoryPostgres,
@@ -37,13 +39,6 @@ import {
   UsdRepositoryCoingecko,
   UsdRepositoryCow,
   UsdRepositoryFallback,
-  cowApiClients,
-  createNewPostgresPool,
-  createTelegramBot,
-  getViemClients,
-  redisClient,
-  ExpiredOrdersRepository,
-  ExpiredOrdersRepositoryPostgres
 } from '@cowprotocol/repositories';
 
 import ms from 'ms';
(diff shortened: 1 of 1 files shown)


apps/api/src/app/routes/__chainId/tokens/__address/balances.ts
alfetopito, line 1: I don't think this is the right route.
Here we want to check account `balances`, but the path used is `tokens`.
It's misleading IMO to use the same path for querying a token usd price (__chainId/tokens/__address/usdPrice) and an account balance (__chainId/tokens/__address/balances).

I suggest instead to use: `__chainId/account/__address/balances` (and in the future, `allowances`)

Feel free to suggest a different path.
Maybe instead of `account`, use `address`?

libs/repositories/src/repos/TokenBalancesRepository/TokenBalancesRepositoryMoralis.ts
```diff
@@ -0,0 +1,101 @@
...
+@injectable()
+export class TokenBalancesRepositoryMoralis implements TokenBalancesRepository {
+  async getTokenBalances({
+    address,
+    chainId,
+  }: TokenBalanceParams): Promise<TokenBalancesResponse> {
+    const network = MORALIS_CLIENT_NETWORK_MAPPING[chainId];
+    if (!network) {
+      throw new Error('Unsupported chain');
+    }
```
alfetopito, line 58: Right now this mapping only has the supported chains.
Would be great to implement it like the usd estimation endpoint, where chains other than supported ones are supported.
This will allow us to enable balance checking for target chains that we don't support as swap chains, such as BTC, Solana etc.

Not on this PR, but do keep it in mind.
Maybe add a TODO so we don't lose the info.

libs/repositories/src/repos/TokenBalancesRepository/TokenBalancesRepositoryMoralis.ts
```diff
@@ -0,0 +1,101 @@
...
+    if (!response.ok) {
+      throw new Error(
+        `Moralis API error: ${response.status} ${response.statusText}`
+      );
+    }
+
+    const asJson = await response.json();
+    if (!isMoralisBalanceResponse(asJson)) {
+      throw new Error('Invalid Moralis response');
+    }
```
alfetopito, line 71: Nitpick: wrap `.json()` call as it can fail? Maybe not needed, just thinking out loud.

alfetopito (commented):
A few nipicks and one main request regarding the new route

shoom3301 (approved):

PR #155, opened 2025-09-03
feat(notifications): add expired orders notifications

shoom3301:
1. Added a new producer `ExpiredOrdersNotificationProducer`
2. The producer periodically (once in ~14sec) checks OrderBook DB and selects expired orders
3. Order is considered as expired when:
 - validTo > lastCheck && validTo <= now
 - AND it's not filled
4. Order is considered as filled when sum of sell/buy amounts from `trades` table is less than order sell/buy amount

### DIFF
--- a/apps/notification-producer/src/producers/expired-orders/getExpiredOrderNotification.ts
+++ b/apps/notification-producer/src/producers/expired-orders/getExpiredOrderNotification.ts
@@ -1,0 +1,51 @@
+import { PushNotification } from '@cowprotocol/notifications';
+import { Erc20Repository, ParsedExpiredOrder } from '@cowprotocol/repositories';
+import { getExplorerUrl } from '@cowprotocol/shared';
+import { type SupportedChainId } from '@cowprotocol/cow-sdk';
+import { getNotificationSummary } from '../../utils/getNotificationSummary';
+
+export interface ExpiredOrderNotificationContext {
+  chainId: SupportedChainId;
+  nowTimestamp: number;
+  lastCheckTimestamp: number;
+  isEthFlowOrder: boolean;
+  owner: string;
+  erc20Repository: Erc20Repository;
+}
+
+export async function getExpiredOrderNotification(
+  expiredOrder: ParsedExpiredOrder,
+  notificationContext: ExpiredOrderNotificationContext
+): Promise<PushNotification> {
+  const { chainId, lastCheckTimestamp, nowTimestamp, isEthFlowOrder, owner, erc20Repository } = notificationContext;
+
+  const summary = await getNotificationSummary({
+    chainId,
+    isEthFlowOrder,
+    erc20Repository,
+    sellAmount: expiredOrder.sellAmount,
+    buyAmount: expiredOrder.buyAmount,
+    sellTokenAddress: expiredOrder.sellTokenAddress,
+    buyTokenAddress: expiredOrder.buyTokenAddress
+  });
+
+  const title = `🕐 Order ${summary} has expired`;
+  const message = `
+  Expiration time: ${new Date(expiredOrder.validTo * 1000).toISOString()}.
+  Account: ${owner}.
+  `.trim();
+
+  const url = getExplorerUrl(chainId, expiredOrder.uid);
+
+  return {
+    id: 'OrderExpired-' + expiredOrder.validTo + '-' + lastCheckTimestamp,
+    account: expiredOrder.owner.toLowerCase(),
+    title,
+    message,
+    url,
+    context: {
+      chainId: chainId.toString(),
+      nowTimestamp: nowTimestamp.toString()
+    }
+  };
+}
--- a/apps/notification-producer/src/utils/getNotificationSummary.ts
+++ b/apps/notification-producer/src/utils/getNotificationSummary.ts
@@ -1,0 +1,38 @@
+import { ALL_SUPPORTED_CHAINS_MAP, SupportedChainId } from '@cowprotocol/cow-sdk';
+import { getAddress } from 'viem';
+import { ChainNames, formatAmount, formatTokenName } from '@cowprotocol/shared';
+import { Erc20Repository } from '@cowprotocol/repositories';
+
+interface OrderInfoForNotificationParams {
+  chainId: SupportedChainId;
+  isEthFlowOrder: boolean;
+  erc20Repository: Erc20Repository;
+  sellTokenAddress: string;
+  buyTokenAddress: string;
+  sellAmount: string | bigint;
+  buyAmount: string | bigint;
+}
+
+export async function getNotificationSummary(params: OrderInfoForNotificationParams): Promise<string> {
+  const { chainId, isEthFlowOrder, erc20Repository } = params;
+
+  const sellToken = isEthFlowOrder
+    ? ALL_SUPPORTED_CHAINS_MAP[chainId].nativeCurrency
+    : await erc20Repository.get(
+      chainId,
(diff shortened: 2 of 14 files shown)


apps/notification-producer/src/producers/expired-orders/getExpiredOrderNotification.ts
```diff
@@ -0,0 +1,51 @@
...
+  const title = `🕐 Order ${summary} has expired`;
+  const message = `
+  Expiration time: ${new Date(expiredOrder.validTo * 1000).toISOString()}.
+  Account: ${owner}.
+  `.trim();
+
+  const url = getExplorerUrl(chainId, expiredOrder.uid);
+
+  return {
+    id: 'OrderExpired-' + expiredOrder.validTo + '-' + lastCheckTimestamp,
```
alfetopito, line 41: Can we add the orderId to the notification id to avoid the issue mentioned below?

apps/notification-producer/src/utils/getNotificationSummary.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export async function getNotificationSummary(params: OrderInfoForNotificationParams): Promise<string> {
+  const { chainId, isEthFlowOrder, erc20Repository } = params;
+
+  const sellToken = isEthFlowOrder
+    ? ALL_SUPPORTED_CHAINS_MAP[chainId].nativeCurrency
+    : await erc20Repository.get(
+      chainId,
+      getAddress(params.sellTokenAddress)
+    );
```
alfetopito, line 24: I mentioned in the other PR, erc20Repository can handle native tokens.

```suggestion
  const sellToken = await erc20Repository.get(
      chainId,
      isEthFlowOrder? NATIVE_TOKEN_ADDRESS : getAddress(params.sellTokenAddress)
    );
```

libs/repositories/src/repos/ExpiredOrdersRepository/ExpiredOrdersRepository.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export interface ExpiredOrder<T = Buffer> {
+  uid: T;
+  owner: T;
+  valid_to: number;
+  sell_token: T;
+  buy_token: T;
+  sell_amount: string;
+  buy_amount: string;
+  kind: 'sell' | 'buy';
```
alfetopito, line 18: Nitpick: extract type for `kind` to reuse here and the type below?

libs/repositories/src/repos/ExpiredOrdersRepository/ExpiredOrdersRepositoryPostgres.ts
```diff
@@ -0,0 +1,69 @@
...
+    const prodDb = getOrderBookDbPool('prod', chainId);
+    const barnDb = getOrderBookDbPool('barn', chainId);
+
+    const prodExpiredOrdersResult = await this.fetchExpiredOrdersFromDb(context, prodDb);
+    const barnExpiredOrdersResult = await this.fetchExpiredOrdersFromDb(context, barnDb);
+
+    const allExpiredOrders = [...(prodExpiredOrdersResult?.rows || []), ...(barnExpiredOrdersResult?.rows || [])];
+
+    const accountsMap = accounts.reduce<Record<string, 1 | undefined>>((acc, account) => {
+      acc[account.toLowerCase()] = 1;
```
alfetopito, line 25: Why storing `1`? If you don't need the value, how about using a Set instead?

libs/repositories/src/repos/ExpiredOrdersRepository/ExpiredOrdersRepositoryPostgres.ts
```diff
@@ -0,0 +1,69 @@
...
+          COALESCE(SUM(buy_amount), 0)  AS filled_buy
+        FROM trades
+        GROUP BY order_uid
+      )
+      SELECT o.uid, o.kind, o.owner, o.valid_to, o.sell_token, o.buy_token, o.sell_amount, o.buy_amount
+      FROM orders o
+      LEFT JOIN trade_sums t ON o.uid = t.order_uid
+      WHERE
+        o.valid_to > ${lastCheckTimestamp}
+        AND o.valid_to <= ${nowTimestamp}
```
alfetopito, line 57: Suggestion: like we do in the UI, what about adding a buffer time, to avoid notifying orders as expired and soon after having them trade?
Add (or remove?) 1min from the nowTimestamp just to prevent them from being notified right away.

libs/repositories/src/repos/ExpiredOrdersRepository/ExpiredOrdersRepositoryPostgres.ts
```diff
@@ -0,0 +1,69 @@
...
+    const { accounts, lastCheckTimestamp, nowTimestamp } = context;
+
+    if (accounts.length === 0) return null;
+
+    const query = `
+      WITH trade_sums AS (
+        SELECT
+          order_uid,
+          COALESCE(SUM(sell_amount), 0) AS filled_sell,
+          COALESCE(SUM(buy_amount), 0)  AS filled_buy
```
alfetopito, line 48: No idea how to do that here, but would be great to have a threshold (0.1%?) to consider partial fills, or almost-100% fills not count as expired

libs/repositories/src/repos/ExpiredOrdersRepository/ExpiredOrdersRepositoryPostgres.ts
```diff
@@ -0,0 +1,69 @@
...
+        FROM trades
+        GROUP BY order_uid
+      )
+      SELECT o.uid, o.kind, o.owner, o.valid_to, o.sell_token, o.buy_token, o.sell_amount, o.buy_amount
+      FROM orders o
+      LEFT JOIN trade_sums t ON o.uid = t.order_uid
+      WHERE
+        o.valid_to > ${lastCheckTimestamp}
```
anxolin, line 56: Just curious, most of the times the SQL engines have a way to inject the parameters as parameters so you avoid SQL injections. Is there a easy way to do it here?

libs/repositories/src/repos/ExpiredOrdersRepository/ExpiredOrdersRepositoryPostgres.ts
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  private async fetchExpiredOrdersFromDb(context: ExpiredOrdersContext, db: Pool): Promise<QueryResult<ExpiredOrder> | null> {
+    const { accounts, lastCheckTimestamp, nowTimestamp } = context;
+
+    if (accounts.length === 0) return null;
+
+    const query = `
+      WITH trade_sums AS (
+        SELECT
```
anxolin, line 45: this do a table scan of the whole trades. Try to filter earlier using a strong filter that is an index. Then you can join by orderId to the trades and do this grouping

shoom3301: I optimized the query, thanks!
https://github.com/cowprotocol/bff/pull/155/commits/62970391b5e9fd6014db1728d6e3ba434e608bd1

libs/repositories/src/repos/ExpiredOrdersRepository/ExpiredOrdersRepositoryPostgres.ts
shoom3301: 1 minute

PR #165, opened 2025-09-04
fix: do not send trade notifications of bridging orders

shoom3301:
1. Updated cow-sdk to v7
2. Created new `OrdersAppDataRepositoryPostgres` to get app-datas of orders. The repo caches the results and does search in both prod and barn envs
3. Added a case to not send trade notifications for bridging orders

### DIFF
--- a/libs/repositories/src/repos/OrdersAppDataRepository/OrdersAppDataRepository.ts
+++ b/libs/repositories/src/repos/OrdersAppDataRepository/OrdersAppDataRepository.ts
@@ -1,0 +1,5 @@
+import { LatestAppDataDocVersion, SupportedChainId } from '@cowprotocol/cow-sdk';
+
+export interface OrdersAppDataRepository {
+  getAppDataForOrders(chainId: SupportedChainId, uids: string[]): Promise<Map<string, LatestAppDataDocVersion>>;
+}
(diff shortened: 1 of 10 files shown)


libs/repositories/src/repos/OrdersAppDataRepository/OrdersAppDataRepository.ts
```diff
@@ -0,0 +1,5 @@
+import { LatestAppDataDocVersion, SupportedChainId } from '@cowprotocol/cow-sdk';
+
+export interface OrdersAppDataRepository {
+  getAppDataForOrders(chainId: SupportedChainId, uids: string[]): Promise<Map<string, LatestAppDataDocVersion>>;
```
alfetopito, line 4: The appData version can be of any version OR not follow that pattern at all.
So, you could use https://github.com/cowprotocol/app-data/blob/main/src/generatedTypes/index.ts#L36 and maybe also an empty obj because we can't tell.

libs/repositories/src/repos/OrdersAppDataRepository/OrdersAppDataRepositoryPostgres.ts
```diff
@@ -0,0 +1,118 @@
...
+  private async fetchAppDataFromDb(uids: string[], db: Pool): Promise<AppDataFromDbResult> {
+    if (!uids.length) return { missingAppDataUids: [], uidToAppData: new Map() };
+
+    const byteaUids = uids.map(hexStringToBytes);
+
+    const query = `
+        SELECT o.uid, a.full_app_data
+        FROM orders o
+          JOIN app_data a ON o.app_data = a.contract_app_data
+        WHERE o.uid = ANY($1) LIMIT ${LIMIT}
```
alfetopito, line 85: Why limit to 100?
Couldn't we limit instead to the size of the list being searched for?
For example, what if we are asking for 101 orders while limiting to 100?

shoom3301: We batch the queries, so it's guaranteed that query will be asked for not more than LIMIT orders

PR #201, opened 2026-02-19
Test exposing origin in unauthorized response

naanselmo:


limitofzero:
@naanselmo is it still relevant?

PR #211, opened 2026-04-14
fix: update sceme validation rules + tenderly types

limitofzero:
Tenderly simulation runs successfully, but when our BFF tries to serialize the response, Fastify's schema validation fails: The value of   '#/items/properties/stateDiff/items/properties/original' does not match schema definition  and it returns 500

## Summary by CodeRabbit

* **Improvements**
  * Simulation bundle endpoint now accepts numbers, booleans, and arrays for state-diff entries in addition to strings, objects, and nulls.

* **Bug Fixes**
  * Preserves falsy values (e.g., 0, false, empty arrays) when normalizing state diffs to avoid misclassifying valid values as missing.

### DIFF
--- a/libs/repositories/src/repos/SimulationRepository/tenderlyTypes.ts
+++ b/libs/repositories/src/repos/SimulationRepository/tenderlyTypes.ts
@@ -578,8 +578,8 @@ interface LogRaw {
 export interface StateDiff {
   address: string;
   soltype: SoltypeElement | null;
-  original: string | Record<string, any> | null;
-  dirty: string | Record<string, any> | null;
+  original: string | number | boolean | Record<string, any> | unknown[] | null;
+  dirty: string | number | boolean | Record<string, any> | unknown[] | null;
   raw?: RawElement[];
 }
 
--- a/libs/repositories/src/utils/buildStateDiff.ts
+++ b/libs/repositories/src/utils/buildStateDiff.ts
@@ -134,7 +134,7 @@ function processRawOnlyDiff(
     if (!updated) {
       const newDiff: StateDiff = {
         address: diff.address,
-        soltype: diff.soltype || null,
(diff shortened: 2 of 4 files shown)


libs/repositories/src/repos/SimulationRepository/tenderlyTypes.ts
```diff
@@ -578,8 +578,8 @@ interface LogRaw {
 export interface StateDiff {
   address: string;
   soltype: SoltypeElement | null;
-  original: string | Record<string, any> | null;
-  dirty: string | Record<string, any> | null;
+  original: string | number | boolean | Record<string, any> | unknown[] | null;
+  dirty: string | number | boolean | Record<string, any> | unknown[] | null;
```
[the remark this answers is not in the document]
limitofzero: fixed. pls check

alfetopito:
@coderabbitai review

libs/repositories/src/utils/buildStateDiff.ts
```diff
@@ -153,7 +153,7 @@ function processSingleDiff(
   diff: StateDiff
 ): StateDiff[] {
   // Handle regular diffs (with complete soltype, original, dirty data)
-  if (diff?.soltype && diff?.original && diff?.dirty) {
+  if (diff?.soltype != null && diff?.original != null && diff?.dirty != null) {
```
alfetopito, line 157: Nitpick: is it intentional to do a loose check instead of strict? (`!=` vs `!==`)

alfetopito (approved):

libs/repositories/src/utils/buildStateDiff.ts
limitofzero: yes, it checks null and undefined at the same moment, I will add comment

shoom3301 (approved):

PR #217, opened 2026-04-29
chore: run yarn lint:fix

kernelwhisperer:
If we want a smaller diff we can turn `semi: true`, but it will be inconsistent with cowswap


Danziger (approved):

limitofzero (approved):
thank you!

PR #243, opened 2026-08-17
fix(notification-producer): logging, Solana bump, address casing

shoom3301:
## What changed
- Bumped `@cowprotocol/cow-sdk` 9.0.0 → 9.2.6, which introduced `SupportedChainId.SOLANA`.
- Notification producer is EVM-only (RPC client, contracts, orderbook DB), so it now uses a dedicated `EVM_CHAIN_IDS` constant (`@cowprotocol/shared`) instead of the general `AllChainIds`. The general `AllChainIds`/`isSupportedChain`/`toSupportedChainId` stay Solana-inclusive, since the repo does support Solana elsewhere (USD prices via CoinGecko, price-impact/slippage estimation) — only the on-chain/settlement-specific code (viem clients, CoW API clients, this producer) excludes it.
- Fixed a real regression from the SDK bump: `coingecko.ts`'s Solana → CoinGecko platform mapping was dropped because Solana moved from `AdditionalTargetChainId` to `SupportedChainId` in the new SDK; restored it under its new enum location so Solana USD prices via CoinGecko keep working.
- `UsdRepositoryCow` (CoW's own settlement-API price source, not the CoinGecko one) explicitly excludes Solana via `isSolanaChain`, since `api.cow.fi` has no Solana network — it would otherwise have thrown on an unhandled `chainIdOrSlug`.
- Replaced the hardcoded `env = 'staging'` in `getTradeNotifications.ts`/`ExpiredOrdersNotificationProducer.ts` with a `COW_PROTOCOL_ENV` env var (`prod`/`staging`, defaults to `prod`) that picks the GPv2Settlement contract and eth-flow address to watch.
- Added step-by-step debug/info logging to `TradeNotificationProducer`, `getTradeNotifications`, and `CmsNotificationProducer` (subscribed accounts, block range being indexed, RPC log-query filter and result count, per-order skip reasons). Previously almost everything was `debug`/silent, so a producer that was polling correctly but had nothing to send was indistinguishable from one that was stuck or misconfigured.
- Standardized address handling on the SDK's `getAddressKey()`/`areAddressesEqual()` instead of ad-hoc `.toLowerCase()` comparisons, and fixed two real bugs found during that audit:
  - `getExpiredOrderNotification`: the notification's `account` used the raw `expiredOrder.owner` instead of the already-resolved owner, so expired-order notifications for eth-flow orders were addressed to the eth-flow contract instead of the actual trader.
  - `CmsNotificationProducer`: the subscription filter compared normalized addresses, but the emitted notification kept the raw, un-normalized `account` from the CMS.
- Documented `COW_PROTOCOL_ENV` in `.env.example`.

## Why
Locally, `nx run notification-producer:start` gave no visibility into whether `TradeNotificationProducer`/`CmsNotificationProducer` were working — after the initial startup log there was nothing at `info` level even when everything was healthy, making "quietly working" indistinguishable from "broken." Debugging that surfaced the SDK's new Solana chain id (which broke the build), a hardcoded staging/prod toggle that should be configurable per deployment, and a couple of address-casing bugs that were misrouting eth-flow notifications.

## QA Testing
Reviewer note: this is a headless queue-processing service (no UI), so there's no browser-based preview flow to test.

Developer verification:
- `nx run <app>:build` for `notification-producer`, `api`, `twap`, and `telegram`, plus the affected `test` targets, all pass locally and in CI (docker builds for all four apps + Test and Lint Affected Projects).
- With `LOG_LEVEL=debug` and `COW_PROTOCOL_ENV=staging` set locally, confirmed `TradeNotificationProducer` logs the full per-poll trail (accounts watched, block range, RPC filter, match/skip reasons) against a real Sepolia trade, and via a direct `eth_getTransactionReceipt` check confirmed the trade's `Trade` event only matches once the producer watches the staging settlement contract — validating the new `COW_PROTOCOL_ENV` toggle against real on-chain data, not just types.

### DIFF
--- a/apps/notification-producer/src/producers/expired-orders/ExpiredOrdersNotificationProducer.ts
+++ b/apps/notification-producer/src/producers/expired-orders/ExpiredOrdersNotificationProducer.ts
@@ -1,4 +1,5 @@
 import {
+  areAddressesEqual,
   BARN_ETH_FLOW_ADDRESSES,
   CowEnv,
   ETH_FLOW_ADDRESSES,
@@ -136,7 +137,7 @@ export class ExpiredOrdersNotificationProducer implements Runnable {
 
       const notifications = await Promise.all(
         expiredOrders.map((order) => {
-          const isEthFlowOrder = ethFlowAddress === getAddressKey(order.owner)
+          const isEthFlowOrder = areAddressesEqual(ethFlowAddress, order.owner)
 
           const orderOwner = isEthFlowOrder
             ? Object.keys(ethFlowOrderOwners).find((key) => {
--- a/apps/notification-producer/src/producers/expired-orders/ExpiredOrdersNotificationProducer.ts
+++ b/apps/notification-producer/src/producers/expired-orders/ExpiredOrdersNotificationProducer.ts
@@ -1,4 +1,4 @@
-import { BARN_ETH_FLOW_ADDRESSES, ETH_FLOW_ADDRESSES, SupportedChainId } from '@cowprotocol/cow-sdk'
+import { BARN_ETH_FLOW_ADDRESSES, ETH_FLOW_ADDRESSES, getAddressKey, SupportedChainId } from '@cowprotocol/cow-sdk'
 import {
   Erc20Repository,
   ExpiredOrdersRepository,
@@ -104,9 +104,7 @@ export class ExpiredOrdersNotificationProducer implements Runnable {
     if (lastCheckTimestampRaw) {
       const lastCheckTimestamp = Number(lastCheckTimestampRaw)
 
-      const ethFlowAddresses = [ETH_FLOW_ADDRESSES[chainId], BARN_ETH_FLOW_ADDRESSES[chainId]].map((t) =>
-        t.toLowerCase()
-      )
+      const ethFlowAddresses = [ETH_FLOW_ADDRESSES[chainId], BARN_ETH_FLOW_ADDRESSES[chainId]].map(getAddressKey)
 
       const accounts = await pushSubscriptionsRepository.getAllSubscribedAccounts()
 
@@ -130,15 +128,16 @@ export class ExpiredOrdersNotificationProducer implements Runnable {
 
       const notifications = await Promise.all(
         expiredOrders.map((order) => {
-          const isEthFlowOrder = ethFlowAddresses.includes(order.owner.toLowerCase())
+          const isEthFlowOrder = ethFlowAddresses.includes(getAddressKey(order.owner))
 
           const orderOwner = isEthFlowOrder
             ? Object.keys(ethFlowOrderOwners).find((key) => {
                 const orderUids = ethFlowOrderOwners[key]
 
+                // order.uid is a 56-byte order digest, not an address, so getAddressKey() would leave it untouched: plain lowercase is correct here.
                 return orderUids.includes(order.uid.toLowerCase())
               })
-            : order.owner.toLowerCase()
+            : getAddressKey(order.owner)
 
           if (!orderOwner) return Promise.resolve(undefined)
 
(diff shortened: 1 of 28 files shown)


apps/notification-producer/src/producers/expired-orders/ExpiredOrdersNotificationProducer.ts
```diff
@@ -130,15 +136,16 @@ export class ExpiredOrdersNotificationProducer implements Runnable {
 
       const notifications = await Promise.all(
         expiredOrders.map((order) => {
-          const isEthFlowOrder = ethFlowAddresses.includes(order.owner.toLowerCase())
+          const isEthFlowOrder = ethFlowAddress === getAddressKey(order.owner)
```
azebuado: `areAddressesEqual`?

apps/notification-producer/src/producers/trade/getTradeNotifications.ts
```diff
@@ -80,7 +99,7 @@ export async function getTradeNotifications(params: GetTradeNotificationParams)
 
     if (log.eventName !== 'Trade') return acc
     if (!orderUid) return acc
-    if (!owner || !ethFlowAddresses.includes(owner.toLowerCase())) return acc
+    if (!owner || ethFlowAddress !== getAddressKey(owner)) return acc
```
azebuado: `areAddressesEqual`

apps/notification-producer/src/producers/trade/getTradeNotifications.ts
```diff
@@ -119,8 +138,9 @@ export async function getTradeNotifications(params: GetTradeNotificationParams)
           break
         }
 
+        // orderUid is a 56-byte order digest, not an address, so getAddressKey() would leave it untouched: plain lowercase is correct here.
         const orderUidLower = orderUid.toLowerCase()
-        const isEthFlowOrder = ethFlowAddresses.includes(owner.toLowerCase())
+        const isEthFlowOrder = ethFlowAddress === getAddressKey(owner)
```
azebuado: `areAddressesEqual`

libs/shared/src/const.ts
```diff
@@ -6,6 +6,12 @@ import {
 } from '@cowprotocol/cow-sdk'
 import { Address } from 'viem'
 
+/**
+ * Chain ids this repo actually supports. All EVM chains in `SupportedChainId`, excluding Solana:
+ * this repo has no non-EVM support (no Solana RPC client, contracts, DB schema, etc.).
+ */
+export type EvmChainId = Exclude<SupportedChainId, SupportedChainId.SOLANA>
```
azebuado: The name/description are not 100% correct.
This repo does support solana for usd prices, price impact estimation etc.
Just for the notification that it doesn't.

Might be better to do it closely to the notification producer as this could affect the rest of the app.

azebuado (changes requested):

libs/shared/src/const.ts
shoom3301: Thanks!
Fixed https://github.com/cowprotocol/bff/pull/243/commits/bb306fb1673b17427facaefc79bd31d0cba937ff

libs/repositories/src/datasources/ethplorer.ts
azebuado, line 1: Nitpick: definitely not the case for this source as they are very limited on chains available, and doesn't matter much as those are not in use, but I expect these data sources to be able to handle Solana. Instead of chaning the record type, I'd say to set the SupportedChainId on Solana to null.

azebuado (approved):
Approved with 1 comment

libs/repositories/src/datasources/ethplorer.ts
shoom3301: Ethplorer doesn't support solana:
https://github.com/EverexIO/Ethplorer/wiki/Ethplorer-API#supported-chains

kernelwhisperer (approved):

PR #245, opened 2026-08-19
feat(notifications): better notifications

azebuado:
# Summary

Improve telegram notifications:

- Order type (from app data)
- Event time
- Partial fills
- Amounts, formatted
- Custom recipient
- Expired

image
image
image
image

# Test

- Unit tests
- Local tests

## Summary by CodeRabbit

- **New Features**
  - Improved trade notifications with clearer order types, status indicators, fill details, timestamps, amounts, and recipient information.
  - Added dedicated expired-order notifications with order-specific messaging and expiration details.
  - Notifications now use the appropriate production or staging order data.
- **Bug Fixes**
  - Improved handling of partial fills, execution updates, missing recipients, and failed block lookups.
  - Telegram order links now use the label “Order details:”.
- **Documentation**
  - Added local notification setup, configuration, testing, and troubleshooting guidance.


azebuado:
@coderabbitai review

apps/notification-producer/src/utils/getBlockTimestamps.ts
```diff
@@ -0,0 +1,12 @@
+interface BlockClient {
+  getBlock: (params: { blockNumber: bigint }) => Promise<{ number: bigint; timestamp: bigint }>
+}
+
+export async function getBlockTimestamps(client: BlockClient, logs: Array<{ blockNumber: bigint | null }>) {
+  const blockNumbers = [
+    ...new Set(logs.map((log) => log.blockNumber).filter((blockNumber): blockNumber is bigint => blockNumber !== null)),
+  ]
+  const blocks = await Promise.all(blockNumbers.map((blockNumber) => client.getBlock({ blockNumber })))
```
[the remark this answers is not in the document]
azebuado: https://github.com/cowprotocol/bff/pull/245/commits/9d880818dc1472dd08af9f17e90ea54cae54092a

apps/notification-producer/src/utils/formatTradeNotification.ts
```diff
@@ -0,0 +1,71 @@
...
+  chainName,
+  tradeAmounts,
+  order,
+  orderAmounts,
+  executedAmounts,
+}: FormatTradeNotificationParams) {
+  const notification = (title: string, message: string) =>
+    formatOrderNotification({ title, message, timestamp, account, recipient, chainName })
+
+  switch (orderTitle) {
```
shoom3301: I would use orderClass instead of orderTitle here since title might be changed in the future and it's not supposed to serve as a key

apps/notification-producer/src/utils/getBlockTimestamps.ts
```diff
@@ -0,0 +1,34 @@
...
+
+interface BlockClient {
+  getBlock: (params: { blockNumber: bigint }) => Promise<{ number: bigint; timestamp: bigint }>
+}
+
+const BLOCK_FETCH_BATCH_SIZE = 10
+
+export async function getBlockTimestamps(client: BlockClient, logs: Array<{ blockNumber: bigint | null }>) {
+  const blockNumbers = [
+    ...new Set(logs.map((log) => log.blockNumber).filter((blockNumber): blockNumber is bigint => blockNumber !== null)),
```
shoom3301, line 9: Nitpick: `blockNumber is bigint => blockNumber !== null` that's not always true...

shoom3301 (commented):
<details>
<summary>⚠️ <strong>AI Review</strong> (Claude Sonnet 5, worked ~12m): dedup drops interim fill notifications</summary>

**Finding:** [BLOCKING] `getLatestTradeLogs` drops all but the last fill per order, even across unrelated blocks in the same processed range

- Location: `apps/notification-producer/src/producers/trade/getTradeNotifications.ts:94` (call site) and `:231-249` (`getLatestTradeLogs`)
- `getLatestTradeLogs` keeps only the highest `(blockNumber, logIndex)` log per `orderUid` and discards every other log for that UID from the batch being processed (`getTradeNotifications.spec.ts:524-530` proves this: a fill at block 10 and a fill at block 11 for the same order collapse into just the block-11 fill).
- `TradeNotificationProducer` processes up to `MAX_BLOCKS_PER_BATCH = 5000` blocks per batch (`TradeNotificationProducer.ts:20`) and always advances `lastBlock` past the whole batch after processing (`TradeNotificationProducer.ts:170` via `processBlocks`), so any fill dropped by the dedup is gone permanently — it is not retried on the next poll.
- Impact: any order with more than one fill inside a single processed range (TWAP parts, repeated partial limit fills) only gets a notification for its last fill in that range. This is not an edge case — it's the normal shape of a catch-up batch after any producer restart/deploy, and directly undercuts this PR's own "TWAP part is filled" / partial-fill notification feature.
- The commit message introducing this (`fix: group events so multiple notifications are not sent for the same block`) suggests the intent was to dedupe literal duplicate events at the *same* block/position, but the implementation collapses fills across different blocks too.

**Suggested fix**
- Dedupe only exact duplicate `(orderUid, blockNumber, logIndex)` positions (true duplicate events), and send one notification per distinct fill otherwise — the new `getOrderExecutionSnapshots` per-position lookup already gives each fill its correct point-in-time executed amount, so there's no need to collapse to a single "final" fill.
- Add a test where the same order fills twice in one batch and both notifications are expected.

<details>

Not a duplicate of the two existing open threads on this PR, both left by the human reviewer and still unresolved:
- `apps/notification-producer/src/utils/formatTradeNotification.ts:34` — switching on the display string `orderTitle` instead of `orderClass` as the routing key; still true in current code (`formatTradeNotification.ts` still `switch (orderTitle)`). Worth resolving before merge since a future copy change would silently fall through to the generic `default` branch.
- `apps/notification-producer/src/utils/getBlockTimestamps.ts:11` — nitpick on the `blockNumber is bigint` type guard; unresolved but non-blocking.

CodeRabbit's two findings on this PR (execution-snapshot staleness, unbounded RPC concurrency in `getBlockTimestamps`) are both resolved and the fixes look correct: `getOrderExecutionSnapshots` computes cumulative amounts as of each trade's exact position, and `getBlockTimestamps` now batches `getBlock` calls 10 at a time with per-block error handling.

</details>

<details>

```text
Verify this finding against current code. Fix only if still valid, keep the change minimal, and validate with targeted tests.

Context:
- apps/notification-producer/src/producers/trade/getTradeNotifications.ts:94, :231 (getLatestTradeLogs)
- Failure mode: when more than one Trade log for the same orderUid appears in one processed block range, all but the highest-position log are discarded and never retried, silently dropping notifications for those fills (most likely to happen after a producer restart processes a large catch-up batch, or for TWAP orders with multiple parts filling close together).
- Expected fix: dedupe only exact duplicate (orderUid, blockNumber, logIndex) positions; keep one notification per genuinely distinct fill using the existing per-position execution snapshot.
- Expected test: two distinct fills for the same order in one batch -> two notifications.
```

</details>

_Generated using the `pr-review` skill from the [CoW Protocol skills repo](https://github.com/cowprotocol/skills)._

</details>

libs/repositories/src/repos/OrdersRepository/OrdersRepositoryPostgres.ts
```diff
@@ -0,0 +1,173 @@
...
+  NotificationOrder,
+  OrderExecutionPosition,
+  OrdersRepository,
+} from './OrdersRepository'
+
+const LIMIT = 100
+
+export class OrdersRepositoryPostgres implements OrdersRepository {
+  async getOrders(chainId: SupportedChainId, uids: string[]): Promise<Map<string, NotificationOrder>> {
+    const prodDb = getOrderBookDbPool('prod', chainId)
```
shoom3301: Since you touched it, can use add `COW_PROTOCOL_ENV` usage here please? So we read only the DB corresponded to COW_PROTOCOL_ENV

libs/repositories/src/repos/OrdersRepository/OrdersRepositoryPostgres.ts
```diff
@@ -0,0 +1,173 @@
...
+    const barnDb = getOrderBookDbPool('barn', chainId)
+    const barnSnapshots = await this.fetchOrderExecutionSnapshotsFromDb(missingPositions, barnDb)
+
+    return new Map([...prodSnapshots, ...barnSnapshots])
+  }
+
+  private async fetchOrdersFromDb(uids: string[], db: Pool): Promise<Map<string, NotificationOrder>> {
+    if (uids.length === 0) return new Map()
+
+    const chunks = chunkArray(uids, LIMIT)
```
shoom3301, line 32: Nice! Thanks for adding that, that what I was worried about

shoom3301 (changes requested):
Would be great to color the notifications with emojis

azebuado:
> ⚠️ **AI Review** (Claude Sonnet 5, worked ~12m): dedup drops interim fill notifications
> 
> **Finding:** [BLOCKING] `getLatestTradeLogs` drops all but the last fill per order, even across unrelated blocks in the same processed range
> 
>     * Location: `apps/notification-producer/src/producers/trade/getTradeNotifications.ts:94` (call site) and `:231-249` (`getLatestTradeLogs`)
> 
>     * `getLatestTradeLogs` keeps only the highest `(blockNumber, logIndex)` log per `orderUid` and discards every other log for that UID from the batch being processed (`getTradeNotifications.spec.ts:524-530` proves this: a fill at block 10 and a fill at block 11 for the same order collapse into just the block-11 fill).
> 
>     * `TradeNotificationProducer` processes up to `MAX_BLOCKS_PER_BATCH = 5000` blocks per batch (`TradeNotificationProducer.ts:20`) and always advances `lastBlock` past the whole batch after processing (`TradeNotificationProducer.ts:170` via `processBlocks`), so any fill dropped by the dedup is gone permanently — it is not retried on the next poll.
> 
>     * Impact: any order with more than one fill inside a single processed range (TWAP parts, repeated partial limit fills) only gets a notification for its last fill in that range. This is not an edge case — it's the normal shape of a catch-up batch after any producer restart/deploy, and directly undercuts this PR's own "TWAP part is filled" / partial-fill notification feature.
> 
>     * The commit message introducing this (`fix: group events so multiple notifications are not sent for the same block`) suggests the intent was to dedupe literal duplicate events at the _same_ block/position, but the implementation collapses fills across different blocks too.

Maybe I didn't phrase the commit msg clearly enough.
This exactly the intended behaviour.

Considering the following:
- Limit order, partially fillable
- In the same block check range (5000 blocks), there are multiple fills
- They get detected at the same time, so notitications would be emitted at the same time
- User would get multiple fill notifications for the same order at the same time

That's quite annoying IMO, so I intentionally chose to collapse all of them into a single notification, for the final state only, not any potential intermediary notifications.

Regarding the other comments, they are valid, will reply to them in the respective comments.

shoom3301 (approved):

shoom3301 (approved):

apps/notification-producer/src/utils/formatOrderNotification.ts
```diff
@@ -0,0 +1,42 @@
...
+    recipient && !areAddressesEqual(account, recipient) ? `\nRecipient: ${abbreviateAddress(recipient)}` : ''
+
+  return {
+    title: `${title} at ${time} UTC`,
+    message: `${message}\n\nAccount: ${abbreviateAddress(account)}${recipientMetadata}\nChain: ${chainName}`,
+  }
+}
+
+function abbreviateAddress(address: string): string {
+  return `${address.slice(0, 6)}...${address.slice(-4)}`
```
limitofzero, line 41: it seemed to me that I've already seen method like this in bw, but I'm not sure

apps/notification-producer/src/utils/getBlockTimestamps.ts
```diff
@@ -0,0 +1,19 @@
...
+  ]
+  const blocks: Array<{ number: bigint; timestamp: bigint }> = []
+
+  for (let index = 0; index < blockNumbers.length; index += BLOCK_FETCH_BATCH_SIZE) {
+    const batch = blockNumbers.slice(index, index + BLOCK_FETCH_BATCH_SIZE)
+    blocks.push(...(await Promise.all(batch.map((blockNumber) => client.getBlock({ blockNumber })))))
```
limitofzero, line 15: is it expected that we don't have retries here?

limitofzero (approved):

apps/notification-producer/src/utils/formatOrderNotification.ts
azebuado: Yes, Sasha told me to duplicate as he had the same on his PR, as we didn't know which one would be merged first.
Will refactor.

apps/notification-producer/src/utils/getBlockTimestamps.ts
azebuado: Good point, we probably should. Will address.

azebuado: Actually, we already have retries at the upper level. Retrying it here would be redundant.

apps/notification-producer/src/utils/formatOrderNotification.ts
azebuado: https://github.com/cowprotocol/bff/pull/247

PR #251, opened 2026-08-26
fix(getUsdPrice): prevent calling cg/cow when not a token

azebuado:
# Summary

Part of https://linear.app/cowswap/issue/FE-449/address-coingecko-fallback-issues-in-bff

Add an existence check in front of the price path. BFF already has a 24h-cached ERC-20 lookup (factories.ts:62-69), so an address with no contract on the requested chain can 404 immediately instead of after two upstream calls. Clears ~84% of our 404s and works without identifying the caller. Highest payoff item.

# Testing

Unit tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved USD price lookups by validating ERC20 token existence before requesting prices.
  * Prevented invalid, nonexistent, or chain-incompatible token addresses from producing misleading price results.
  * Preserved fallback behavior for native tokens, unsupported chains, non-EVM assets, and temporary RPC failures.
* **Tests**
  * Added coverage for valid, missing, and malformed tokens, unsupported chains, omitted addresses, cross-chain mismatches, and RPC errors.

### DIFF
--- a/libs/repositories/src/repos/UsdRepository/UsdRepositoryFallback.ts
+++ b/libs/repositories/src/repos/UsdRepository/UsdRepositoryFallback.ts
@@ -1,4 +1,4 @@
-import { BTC_CURRENCY_ADDRESS, isSolanaChain, isSupportedChain } from '@cowprotocol/cow-sdk'
+import { BTC_CURRENCY_ADDRESS, isNonEvmChain, isSupportedChain } from '@cowprotocol/cow-sdk'
 import { logger } from '@cowprotocol/shared'
 import { base58 } from '@scure/base'
 import { injectable } from 'inversify'
@@ -99,7 +99,7 @@ export class UsdRepositoryFallback implements UsdRepository {
 
     // Erc20Repository only has RPC clients for EVM chains with CoW Protocol settlement, so there is
     // nothing to check against for Solana, Bitcoin or Optimism. Same guard as UsdRepositoryCow.
-    if (!chainId || !isSupportedChain(chainId) || isSolanaChain(chainId)) {
+    if (!chainId || !isSupportedChain(chainId) || isNonEvmChain(chainId)) {
       return true
     }
 
--- a/libs/repositories/src/repos/UsdRepository/UsdRepositoryFallback.ts
+++ b/libs/repositories/src/repos/UsdRepository/UsdRepositoryFallback.ts
@@ -1,8 +1,10 @@
-import { BTC_CURRENCY_ADDRESS } from '@cowprotocol/cow-sdk'
+import { BTC_CURRENCY_ADDRESS, isSolanaChain, isSupportedChain } from '@cowprotocol/cow-sdk'
 import { logger } from '@cowprotocol/shared'
 import { base58 } from '@scure/base'
 import { injectable } from 'inversify'
 import { isAddress } from 'viem'
+import { getSupportedCoingeckoChainId } from '../../utils/coingeckoUtils'
+import { Erc20Repository } from '../Erc20Repository/Erc20Repository'
 import { PricePoint, PriceStrategy, UsdRepository } from './UsdRepository'
 
 function isValidTokenAddress(tokenAddress: string | undefined): boolean {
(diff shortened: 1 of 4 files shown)


libs/repositories/src/repos/UsdRepository/UsdRepositoryFallback.ts
```diff
@@ -74,4 +84,42 @@ export class UsdRepositoryFallback implements UsdRepository {
...
+    const chainId = getSupportedCoingeckoChainId(chainIdOrSlug)
+
+    // Erc20Repository only has RPC clients for EVM chains with CoW Protocol settlement, so there is
+    // nothing to check against for Solana, Bitcoin or Optimism. Same guard as UsdRepositoryCow.
+    if (!chainId || !isSupportedChain(chainId) || isSolanaChain(chainId)) {
+      return true
+    }
+
+    try {
+      const erc20 = await this.erc20Repository.get(chainId, tokenAddress)
```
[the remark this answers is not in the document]
azebuado: Fixed on https://github.com/cowprotocol/bff/pull/251/commits/920ef63ddb070526d30a164d2d3a71d379e09235

libs/repositories/src/repos/UsdRepository/UsdRepositoryFallback.ts
```diff
@@ -74,4 +84,48 @@ export class UsdRepositoryFallback implements UsdRepository {
...
+  private async tokenExists(chainIdOrSlug: string, tokenAddress: string | undefined): Promise<boolean> {
+    if (!tokenAddress) {
+      return true
+    }
+
+    const chainId = getSupportedCoingeckoChainId(chainIdOrSlug)
+
+    // Erc20Repository only has RPC clients for EVM chains with CoW Protocol settlement, so there is
+    // nothing to check against for Solana, Bitcoin or Optimism. Same guard as UsdRepositoryCow.
+    if (!chainId || !isSupportedChain(chainId) || isSolanaChain(chainId)) {
```
shoom3301: Nitpick: It's better to use `isNonEvmChain` instead of `isSolanaChain` to cover Bitcoin and any other chains

libs/repositories/src/repos/UsdRepository/UsdRepositoryFallback.ts
```diff
@@ -74,4 +84,48 @@ export class UsdRepositoryFallback implements UsdRepository {
     }
     return null
   }
+
+  /**
+   * An address with no ERC20 contract on the requested chain can't be priced by any source, so we
+   * skip both upstream calls instead of paying for them before returning null anyway.
+   * This is most of the wrong-chain traffic we serve (tokens requested on a chain they don't exist on).
+   */
+  private async tokenExists(chainIdOrSlug: string, tokenAddress: string | undefined): Promise<boolean> {
```
shoom3301, line 93: Nitpick: maybe we should wrap `tokenExists` with a caching layer (like a in-memory Set)

shoom3301 (approved):

libs/repositories/src/repos/UsdRepository/UsdRepositoryFallback.ts
azebuado: We could use redis instead of in memory, but if I'm not mistaken, the fact that the token doesn't exist is cached for 24h so it effectively means the same thing. Let me double check that.

azebuado: Yep, token details are already cached, for 24h https://github.com/cowprotocol/bff/blob/6406663a81733f42bd1248972175347e3105548a/libs/services/src/factories.ts#L68

libs/repositories/src/repos/UsdRepository/UsdRepositoryFallback.ts
azebuado: https://github.com/cowprotocol/bff/pull/251/commits/97da76e204ad6caf623d38f6a2f5c7a9f4ab1998

PR #242, opened 2026-08-12
fix: optimize coingecko calls

azebuado:
# Summary

Addressing 3 problems we currently have making we waste a lot of coingecko calls

1. Sepolia. Due to a bug, we are currently calling coingecko when searching for Sepolia usd prices, but they don't offer that.
2. Malformed token addresses. Due to a very permissive model, things like `test` are being queried against coingecko. It should be rejected instead.
3. Native token queries. We do _a lot_ of requests to coingecko using the native token address (0xeeee...), which them all fail.

Addresses part of https://linear.app/cowswap/issue/FE-449/address-coingecko-fallback-issues-in-bff

# Testing

Unit tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved USD price lookups for Ethereum, Solana, and Bitcoin token addresses.
  * Invalid token addresses now safely return no price instead of triggering unnecessary lookups.
  * Price retrieval continues to work when a token address is not provided.
  * Improved recognition of native currencies across supported networks.
  * Corrected chain and platform resolution for CoinGecko pricing, including Sepolia and unmapped chain IDs.

PR #252, opened 2026-08-27
fix(coingecko): properly map coingecko platforms

azebuado:
# Summary

Addresses another part of https://linear.app/cowswap/issue/FE-449/address-coingecko-fallback-issues-in-bff

Fix the platform mapping and native coin id resolution. Higher severity than first recorded: this is not only missing prices, it is wrong prices served with a 200. Three parts:

1. Native lookups pass the platform id where a coin id belongs. For Base, Linea and Ink an unrelated token owns that platform id as its coin id, so its price is returned as the chain's native price. Base resolves to $0.00000083 instead of ETH at $2,449. Regression from #242, verified twice against CoinGecko's live API.

2. BNB, Polygon, Avalanche, Arbitrum and Optimism miss and fall back to Cow. This was the originally recorded symptom.

3. Unmapped chain ids and unknown slugs fall through to the raw input, which 404s every time.

Approach: adopt CoinGecko's own /asset_platforms feed, which publishes chain_identifier, platform id and native_coin_id together, as a committed generated file refreshed by hand. Nothing fetches at build, install or CI time. It is a strict superset of our 195-entry hand-written table, so that table is deleted rather than duplicated.

# Testing

- Unit tests
- e2e coingecko tests ran locally

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved native cryptocurrency price lookups for current and historical prices across supported networks.
  * Preserved contract-address pricing for ERC-20 tokens.
  * Unknown or unsupported network identifiers now return no result instead of being misinterpreted.

* **Improvements**
  * Expanded platform coverage using up-to-date CoinGecko data.
  * Improved mapping between networks and their native currencies.
  * Added tooling to refresh CoinGecko platform information.


azebuado:
@coderabbitai review

shoom3301 (approved):

PR #254, opened 2026-08-27
fix(getUsdPrice): don't stop of first failure

azebuado:
# Summary

Another part of https://linear.app/cowswap/issue/FE-449/address-coingecko-fallback-issues-in-bff

Don't return a 500 on first 500 found. Try all repositories instead as one might have a price.

## Details

Make fallback work on upstream errors, not just null, and stop caching upstream errors as "no price". Implemented together, since neither works alone: fixing only the fallback is inert because Coingecko failures never reach it as throws, and fixing only Coingecko would turn its errors straight into 500s.

- UsdRepositoryFallback had a bare await per repository, so a throw aborted the loop and escaped as a 500 instead of trying the next source. A throw is now treated like a null and the next source is tried.

- Both Coingecko response handlers short-circuited before throwIfUnsuccessful: a 429 or 5xx carries no price field, matched the !data?.[key]?.usd condition and returned null. That null was cached for the 30 minute NULL TTL, hid the outage behind a 404 long after recovery, and made the fallback log line count failures as unknown tokens. Only a 404 is now treated as an answer.

- When no price is found and any source failed, the error is rethrown rather than reported as no price. Reason below in the notes: a 404 makes the frontend blacklist the token for the whole session, so an outage must never look like one.

- UsdRepositoryCache needed no change, since cacheValue already sits after the await and a throw skips it. That was implicit and is now asserted in tests.

- Watch after deploy: this deliberately converts a class of silent 404s into 5xx. Volume is unknown precisely because these failures have been recorded as nulls until now.

# Testing

- Unit tests
- Ran coingecko e2e locally

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved USD price handling when upstream pricing services experience errors.
  - Invalid or unsuccessful responses now surface as errors instead of being incorrectly treated as unavailable prices.
  - Confirmed that failed price lookups are not cached as missing values.
  - Fallback pricing sources are now tried more reliably, with failures recorded and successful results returned when available.
  - Missing prices, unavailable tokens, zero-valued prices, and not-found responses continue to be handled appropriately.

### DIFF
--- a/libs/repositories/src/repos/UsdRepository/UsdRepositoryCache.spec.ts
+++ b/libs/repositories/src/repos/UsdRepository/UsdRepositoryCache.spec.ts
@@ -158,7 +158,11 @@ describe('UsdRepositoryCache', () => {
       const pricePromise = usdRepositoryCache.getUsdPrice(chainId, WETH)
 
       // THEN: The call throws an awful error
-      expect(pricePromise).rejects.toThrow('💥 Booom!')
+      await expect(pricePromise).rejects.toThrow('💥 Booom!')
+
+      // THEN: The failure is NOT cached as "no price". Caching it would hide the outage behind a
+      // 404 for the whole NULL TTL, long after the upstream recovered.
+      expect(redisMock.set).not.toHaveBeenCalled()
     })
   })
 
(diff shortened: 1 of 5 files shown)


azebuado:
@coderabbitai review

shoom3301 (approved):

PR #255, opened 2026-08-27
fix(api): improve logging

azebuado:
# Summary

Overhaul the logging of the api app:
- All logs now have a req-id
- Add http timeouts to external services
- Downgrade common logs (price and slippage) to debug
- Log all proxy error responses
- Log all tokens proxy responses
- Censor sensitive headers from logs
- Cache upstream errors briefly to avoid hammering services too hard
- Prevent retry barrage on failure (11 calls sequentially upon 500s)

# Testing

- Unit tests
- New e2e file ran manually
- Started the server locally and tested the tokens proxy
image

## Summary by CodeRabbit

* **Reliability**
  * Added upstream timeouts and improved handling of transport failures, interrupted responses, and repeated failed requests.
  * Temporarily caches upstream failures to reduce disruption and unnecessary retries.
  * Improved fallback behavior when token or price data is unavailable.

* **Observability**
  * Added request identifiers and richer response logging, including status, size, content type, and selected response details.
  * Improved outbound service-call diagnostics while reducing routine proxy log noise.
  * Sensitive authorization, cookie, and API-key data is excluded from logs.

* **Bug Fixes**
  * Improved handling of upstream errors and timeout conditions across supported data sources.


azebuado:
@coderabbitai review

apps/api/src/utils/registerProxy.ts
```diff
@@ -0,0 +1,185 @@
...
+      },
+    } as ReplyOptions,
+  })
+}
+
+// Only ever read a symbol off the request, so they stay clear of Fastify's request generics, which
+// differ between the plugin hooks and the route handler.
+function markStart(request: unknown): void {
+  ;(request as Record<symbol, number>)[START_TIME] = Date.now()
+}
```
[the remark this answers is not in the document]
azebuado: Not an issue, added by prettier.

apps/api/src/utils/logResponseBody.ts
```diff
@@ -0,0 +1,35 @@
...
+  return policy === 'always' || (policy === 'errors' && status >= 400)
+}
+
+/** Truncates to the cap, reporting whether anything was dropped. */
+export function truncateBody(body: string): { snippet: string; truncated: boolean } {
+  const buffer = Buffer.from(body)
+
+  return buffer.length <= MAX_LOGGED_BODY_BYTES
+    ? { snippet: body, truncated: false }
+    : { snippet: buffer.subarray(0, MAX_LOGGED_BODY_BYTES).toString('utf8'), truncated: true }
```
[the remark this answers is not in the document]
azebuado: Probably not worth it.

libs/repositories/src/datasources/coingecko.ts
```diff
@@ -80,8 +82,11 @@ export function getCoingeckoProClient(apiKey = process.env.COINGECKO_API_KEY): C
     headers: {
       'x-cg-pro-api-key': apiKey,
     },
+    fetch: fetchWithTimeout(),
   })
```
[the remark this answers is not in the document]
azebuado: Fixed

apps/api/src/utils/registerProxy.ts
```diff
@@ -0,0 +1,195 @@
...
+              bodyTruncated: snippet !== undefined && truncated ? true : undefined,
+            },
+            `Proxied to ${name}`
+          )
+        })
+
+        // pipeline rather than body.pipe: pipe does not forward a source error to the destination, so
+        // an upstream body that fails mid-transfer would leave the tap alive, never report, and the
+        // call would go unlogged. Partial responses are exactly what the logging is for.
+        pipeline(body as unknown as Readable, tap, () => undefined)
```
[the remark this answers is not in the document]
azebuado: Fixed again https://github.com/cowprotocol/bff/pull/255/commits/c3277214af5695d992daad8497c9d1816a0b194c

apps/api/src/app/routes/proxies/tokens/index.ts
```diff
@@ -1,24 +1,24 @@
...
-          Host: fastify.config.PROXY_HOST,
-        }
-      },
+      rewriteRequestHeaders: (originalRequest, headers) => ({
+        ...headers,
+        Origin: fastify.config.PROXY_ORIGIN,
+        Host: fastify.config.PROXY_HOST,
+      }),
     },
+    logResponseBody: 'always',
```
azebuado, line 21: Will now log all responses from this endpoint

libs/shared/src/utils/requestContext.ts
```diff
@@ -0,0 +1,24 @@
+import { AsyncLocalStorage } from 'async_hooks'
+
+export interface RequestContext {
+  /** Fastify's per-request id, so a log line can be joined to the request that produced it */
+  reqId: string
```
azebuado, line 5: Reqid logging

apps/api/src/app/plugins/requestContext.ts
azebuado, line 1: <details>
<summary>⚠️ <strong>AI Review</strong> (Claude Opus 5 (1M context), worked 5m): <code>registerProxy</code> silently discards a caller's <code>preHandler</code></summary>

**Finding:** [NON-BLOCKING] `registerProxy` accepts `preHandler` in its type but never runs it

- Location: `apps/api/src/utils/registerProxy.ts:120-139`
- `...options` is spread first, then `preHandler` is assigned unconditionally. Later key wins, so any `preHandler` a caller passes is dropped with no error and no log.
- It type-checks, because the signature is `ProxyDefinition & Omit<FastifyHttpProxyOptions, 'upstream'>` and `preHandler` survives that `Omit`.

This stands out because every other option in the same object composes deliberately:

| option | behaviour |
|---|---|
| `undici` | `{ headersTimeout, bodyTimeout, ...options.undici }` — composes |
| `replyOptions.onResponse` | saved as `callerOnResponse`, invoked at line 205 — composes |
| `replyOptions.onError` | `options.replyOptions?.onError` invoked at line 213 — composes |
| `preHandler` | **silently overwritten** |

The trap is live rather than theoretical: this PR removes a `preHandler` from both `proxies/coingecko/index.ts` and `proxies/socket/index.ts`, so it was in use at both call sites immediately before this change. Re-adding one — an auth check, a rewrite, a per-proxy log — would look correct, type-check, and do nothing.

**Suggested fix**
- Compose it like the reply hooks: keep `options.preHandler`, and after the block check and `markStart(request)`, call it if present.
- Or, if proxies should never carry their own `preHandler`, `Omit` it from the accepted type so the compiler rejects it instead of the runtime ignoring it.
- Worth a line in `registerProxy.spec.ts` asserting a caller-supplied `preHandler` runs, since the current suite would not catch the regression.

<details>

Second pass over `fix/improve-logging` @ `35e782f`, now rebased onto `main` (#251 and #254 merged, so the diff dropped from 15 to 17 commits against a new base). This is a self-review — I wrote most of the code under review, so findings come from running it rather than re-reading my own intent.

Checked and cleared this pass:
- **`redact` coverage of the new logging.** `main.ts:10` passes the shared pino instance to Fastify, so `request.log` is a child. Confirmed with a targeted script that children inherit root `redact`, so the `Authorization`/`Cookie`/`x-api-key` paths cover the new `request.log` lines in `registerProxy` and `responseLogging`, not just repository logs.
- **Both fixes from the previous pass landed** (`35e782f`): the `responseLogging` doc comment now states the real double-send condition, and `registerProxy` imports `MAX_LOGGED_BODY_BYTES` / `LOGGABLE_CONTENT_TYPE` / `LogResponseBody` / `shouldLogBody` from `logResponseBody.ts` instead of redefining them. Each symbol now has one definition.
- **No duplicate log line per proxied request.** `registerProxy.onResponse` sends a stream, and `responseLogging` skips non-string payloads.
- **Only one CoinGecko client exists** (Pro, `coingecko.ts:80`), so there is no second client missing `fetchWithTimeout`/`upstreamLogging`.
- **`undici` composition** preserves the coingecko proxy's `strictContentLength: false` alongside the new timeouts.

Existing threads, not repeated as fresh findings:
- [`coingecko.ts:89` — timing entries retained after transport failures](https://github.com/cowprotocol/bff/pull/255#discussion_r3881456385) (CodeRabbit, Major, unanswered): I read this as a false positive. It describes `startedAt` as keyed by request id needing explicit cleanup, but `upstreamLogging.ts:11` is a `WeakMap<Request, number>`, so a timed-out or aborted call's entry is collected with its `Request` and no `onError` hook is needed. Worth a short reply closing it rather than leaving a Major thread open.
- [`tapBody.spec.ts:12` — fixed sleeps](https://github.com/cowprotocol/bff/pull/255#discussion_r3882172764) (CodeRabbit, Minor, unanswered): real but low-materiality. The 50 ms waits cover in-memory stream work that finishes in microseconds; the suite passed 5/5 with the cache disabled. Awaiting `stream.finished` would be strictly better if the file is touched again.
- Already resolved and verified in current code: `pipeline` instead of `.pipe` for mid-stream failures, `tokens-manual-test` proxy-name isolation, the prettier semicolon, and the UTF-8 truncation nit.

Automated checks run against this head: `yarn lint` (0 errors, 7 pre-existing warnings), `yarn build`, `yarn test` (8 projects green), plus `tapBody` five times with `--skip-nx-cache`.

</details>

<details>

```text
In apps/api/src/utils/registerProxy.ts, the object passed to fastify.register(httpProxy, ...)
spreads `...options` first and then assigns `preHandler` unconditionally. JS object-literal
semantics mean the later key wins, so a caller-supplied `preHandler` is silently discarded.
The type `ProxyDefinition & Omit<FastifyHttpProxyOptions, 'upstream'>` still accepts it, so
this fails silently at runtime rather than at compile time.

Note that the same object composes every other caller option on purpose: `undici` spreads
`...options.undici`, and replyOptions.onResponse / onError are captured and invoked. Only
preHandler is dropped.

Fix by composing it, matching the existing style: keep a reference to options.preHandler and
invoke it inside the wrapper after the failure-block check and markStart(request), preserving
the early return when the block check sends a 503. Alternatively, add 'preHandler' to the Omit
so the compiler rejects it.

Add a test in apps/api/src/utils/registerProxy.spec.ts asserting a caller-supplied preHandler
runs on a forwarded request; the current suite does not cover it.
```

</details>

*Generated using the `pr-review` skill from the [CoW Protocol skills repo](https://github.com/cowprotocol/skills).*

</details>

azebuado: Addressed on https://github.com/cowprotocol/bff/pull/255/commits/0f2a8bcffa8a20882eb8c7dd210f0647bd65a224

shoom3301 (commented):
<details>
<summary>⚠️ <strong>AI Review</strong> (Claude Sonnet 5, worked 10m): follow-up still reproducible on the mid-stream failure logging thread</summary>

**Rechecked**
- Prior thread: [`registerProxy.ts` — mid-stream upstream failure produces no proxy log line](https://github.com/cowprotocol/bff/pull/255#discussion_r3882483637). CodeRabbit's last reply on it already said the finding remained open after commit `c327721`; I independently re-derived the same conclusion from current code rather than taking that reply at face value.
- Code path: `apps/api/src/utils/registerProxy.ts` `onResponse` (the `pipeline(body, tap, (error) => { if (error) recordFailure(...) })` block) and `apps/api/src/utils/tapBody.ts` (`tap.on('close', report)` / `tap.on('error', report)`).
- Test coverage: `apps/api/src/utils/registerProxy.spec.ts` — *"records a transport failure when the upstream body dies mid-transfer"* only asserts `Proxy to ${proxyName} failed` is present; it never asserts `Proxied to ${proxyName}` is absent.

**Result:** Still broken. `pipeline`'s destroy-on-error also drives `tap`'s own `'error'`/`'close'` listeners, so `report()` (and therefore the `Proxied to ${name}` info-level "success" log with a `status`/`bytes` line) fires unconditionally, in addition to the `Proxy to ${name} failed` error log from `recordFailure`. A mid-stream drop currently produces *both* a success-shaped and a failure-shaped log line for the same request, which is exactly the ambiguity this PR's `onResponse`/`recordFailure` split was meant to remove.

**Suggested next step**
- Make the `Proxied to` log conditional on the pipeline completing without an error — e.g. have `tap`'s `onDone` receive/check the pipeline outcome, or have the `pipeline` callback suppress/skip the completion log when `error` is set (order between `tap`'s close/error handlers and the `pipeline` callback isn't guaranteed either way).
- Extend the existing "records a transport failure when the upstream body dies mid-transfer" test to assert `Proxied to ${proxyName}` is **not** logged, since the current test only checks the failure log's presence.

<details>

```text
Verify this follow-up finding against current code only. Confirm whether the mid-stream
upstream-failure logging in apps/api/src/utils/registerProxy.ts still emits a "Proxied to
${name}" success log alongside the "Proxy to ${name} failed" error log for the same request.

Context:
- apps/api/src/utils/registerProxy.ts: onResponse's pipeline(body, tap, callback) calls
  recordFailure(request, error) on a pipeline error.
- apps/api/src/utils/tapBody.ts: tap.on('close', report) and tap.on('error', report) call
  onDone() unconditionally, which is what emits "Proxied to ${name}" from registerProxy.ts.
  pipeline's destroy-on-error triggers tap's 'error'/'close' independently of the pipeline
  callback, so report() runs regardless of whether the transfer failed.
- Test: apps/api/src/utils/registerProxy.spec.ts "records a transport failure when the upstream
  body dies mid-transfer" only asserts the failure log exists; it does not assert the success
  log is absent.

Expected fix: make the "Proxied to ${name}" completion log conditional on the pipeline finishing
without an error (do not assume ordering between tap's close/error listeners and the pipeline
callback). Extend the existing test to assert "Proxied to ${proxyName}" is NOT logged for the
mid-transfer-failure case.
```

</details>

<details>

This PR already carries a thorough self-review pass (see the unresolved thread on `apps/api/src/app/plugins/requestContext.ts` from a prior AI review) plus extensive CodeRabbit coverage. Checked against current code and not repeated as fresh findings:

- `preHandler` silently discarded by `registerProxy` — fixed: the accepted type now `Omit`s `preHandler` (`registerProxy.ts`), with a doc comment explaining why, and no call site passes one.
- `fetchWithTimeout` discarding the caller's `AbortSignal` — fixed: it now combines `input.signal`/`init.signal` with the timeout signal via `AbortSignal.any`.
- `upstreamLogging.ts` `startedAt` map retaining entries after a timeout/abort — verified as a false positive raised by CodeRabbit: `startedAt` is a `WeakMap<Request, number>`, so a request that never reaches `onResponse` is reclaimed with its `Request` object, no explicit cleanup needed.
- `tapBody.spec.ts` fixed-sleep timing (CodeRabbit, Minor) — real but low materiality (in-memory stream work, suite passes consistently); not worth blocking on.
- UTF-8 truncation boundary in `logResponseBody.ts` and the ESLint semicolon nit — both explicitly declined by the author and accepted by CodeRabbit as not worth the added complexity / not an issue.

</details>

_Generated using the `pr-review` skill from the [CoW Protocol skills repo](https://github.com/cowprotocol/skills)._

</details>

libs/shared/src/utils/requestContext.ts
```diff
@@ -0,0 +1,24 @@
+import { AsyncLocalStorage } from 'async_hooks'
```
shoom3301, line 1: Nitpick:

```suggestion
import { AsyncLocalStorage } from 'node:async_hooks'
```

shoom3301: https://nodejs.org/api/async_context.html#class-asynclocalstorage

apps/api/src/utils/registerProxy.ts
```diff
@@ -0,0 +1,235 @@
...
+              url: request.url,
+              method: request.method,
+              status,
+              ms: elapsed(request),
+              contentType: contentType || undefined,
+              bytes,
+              body: snippet,
+              bodyTruncated: snippet !== undefined && truncated ? true : undefined,
+            },
+            `Proxied to ${name}`
```
shoom3301: https://github.com/cowprotocol/bff/pull/255#pullrequestreview-5102903854

>Proxied to ${name} info-level "success" log with a status/bytes line) fires unconditionally, in addition to the Proxy to ${name} failed error log from recordFailure

@azebuado 
I don't think it's too bad, just noticed in the AI review.
Anyway, I have another question. Do we really need to log every successful proxy response? It looks like there is gonna be tons of them.

azebuado: Yes, it might be much. But this is not used a lot currently.
The issue is that some proxies don't follow REST rules. A failure is a 200 with the error in the response.
Without having to create special rules for those, I'd like to log them for inspecting the errors for awhile.
We can turn them off later.

shoom3301 (approved):

apps/api/src/utils/registerProxy.ts
```diff
@@ -0,0 +1,251 @@
...
+        const isReadable = LOGGABLE_CONTENT_TYPE.test(contentType)
+
+        // Logged once the body has flushed, so `bytes` and `ms` cover the whole upstream leg rather
+        // than just its headers.
+        const tap = tapBody(
+          wantsBody && isReadable ? MAX_LOGGED_BODY_BYTES : 0,
+          ({ bytes, snippet, truncated, error }) => {
+            // A body that died in transfer is reported by recordFailure below. Logging it here too
+            // produced a success-shaped 'Proxied to' line, status and all, beside the failure line.
+            if (error) return
```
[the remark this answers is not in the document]
azebuado: Fixed on https://github.com/cowprotocol/bff/pull/255/commits/6f69271bc7eaa0aae8710bf20838caa697b0ba0e

apps/api/src/app/plugins/requestContext.ts
azebuado: <details>
<summary>✅ <strong>AI Review</strong> (Claude Opus 5 (1M context), worked 4m): all prior findings verified fixed on <code>6f69271</code>; no new defects</summary>

Review completed against the pushed head. Every finding raised on this PR is fixed in code and pinned by a test that fails without it. Two threads are still open, but neither is a defect — both need a reply rather than a change.

**Rechecked — each fix confirmed present, and each regression test confirmed non-vacuous**

| Finding | Fix on `6f69271` | Guarded by |
|---|---|---|
| Failure tally was cumulative, not consecutive | `cacheRepository.take(failureCountKey)` clears on a completed response | *forgets earlier failures once the upstream answers again* |
| Client aborts counted as upstream failures | `if (reply.raw.destroyed) return` before `recordFailure` | *does not count a client walking away as an upstream failure* |
| Mid-stream failure logged a success line | `tapBody` carries the transfer error; `onDone` returns early | *records a transport failure when the upstream body dies mid-transfer* |
| `preHandler` silently discarded | omitted from the accepted type → `TS2353` | compile-time |
| Accidental `PROBES` block broke the suite | removed in `16bcf76` | suite compiles, 17/17 |

For each of the first three I removed the fix and confirmed **only** that finding's test fails — in particular *"stops forwarding once failures are sustained"* still passes with the tally-reset removed, so genuine sustained failures still trip the breaker.

**Checks on this head**
- `yarn lint` 0 errors, `yarn build` clean, `yarn test` green across 8 projects on three consecutive runs.
- `registerProxy.spec.ts` run five times: 17/17 every time. Three of those tests drive real sockets and timers, so I checked for flakiness rather than assuming.
- `cacheRepository.take` is `GETDEL` (Redis 6.2+). Confirmed it is not a silent no-op in production: the deployment pins `redis:7.2.5`, and `take` is already used by the telegram connect-token flow.

<details>

Reviewed `fix/improve-logging` @ `6f69271` against `main` (29 files, +1526 −61). Self-review: I wrote most of this code, so conclusions come from running it.

Also checked this pass and found correct:
- `UsdRepositoryCoingecko.ts` — the only production file I had not previously scrutinised. `resolvePlatform` / `resolveNativeCoinId` wrap the existing helpers with an `info` line and return the same values, so the early-exit paths are unchanged in behaviour.
- Empty body (204 / HEAD) and chunked bodies, exercised directly against the built `tapBody` outside the repo: `bytes: 0` with no snippet and no spurious truncation flag; chunks reassembled correctly.
- Breaker recovery arithmetic: the count key (10s) expires during the block (20s), so recovery starts from zero rather than re-blocking on the next single failure.

One consequence worth knowing, deliberate rather than a defect: a **client abort also clears the failure tally**, because the tap closes without an error and reaches the same success branch. It is defensible — the upstream had already sent headers and bytes, which is what the tally measures — and it errs toward forwarding rather than blocking, which is the safer direction for this particular breaker. Flagging it so the behaviour is not a surprise later.

Open, needing a reply rather than a code change:
- [`requestContext.ts:1` — `node:async_hooks` prefix](https://github.com/cowprotocol/bff/pull/255#discussion_r3925354327): applying it here would make this the only `node:`-prefixed builtin import in the repo — `'path'`, `'stream'`, `'fs'`, `'http'`, `'http2'`, `'crypto'` are all unprefixed. Better repo-wide than as a single-file exception.
- [`coingecko.ts:89` — timing entries retained after transport failures](https://github.com/cowprotocol/bff/pull/255#discussion_r3881456385) (CodeRabbit, Major, still unanswered): a false positive. `upstreamLogging.ts:11` is a `WeakMap<Request, number>`, so an entry for a call that never reaches `onResponse` is reclaimed with its `Request`; no `onError` hook is needed.

Verification throughout is automated: targeted and full test runs, fix-removal checks, and inspection of the built output. No manual or browser QA applies to this change.

</details>

*Generated using the `pr-review` skill from the [CoW Protocol skills repo](https://github.com/cowprotocol/skills).*

</details>

PR #256, opened 2026-08-31
feat: price history endpoint & supply endpoint 

kernelwhisperer:
## Summary

Add BFF endpoints for price history and token supply.

- `GET /{chainId}/tokens/{tokenAddress}/priceHistory`
- `GET /{chainId}/tokens/{tokenAddress}/supply`
- Use the upstream provider first and Codex as fallback.
- Add unit and live integration tests. **~750 LoC**
- Add response caching.
  - Price history has no server-side response cache. It only returns a 30-second HTTP cache header.
  - Token supply has a one-hour in-memory BFF cache and a one-hour HTTP cache header.

`PRICE_HISTORY_PROVIDER_ORDER=1,2`

## To Test

Run it locally together with https://github.com/cowprotocol/cowswap/pull/8009

Usage: https://github.com/cowprotocol/cowswap/blob/feat/price-chart/apps/cowswap-frontend/src/modules/priceChart/api/fetchPriceChartData.ts

## Summary by CodeRabbit

- **New Features**
  - Added token price-history API support with configurable providers, multiple intervals, validation, caching, and automatic fallback.
  - Added Codex and upstream data-provider integrations for market data.
  - Added token supply retrieval with one-hour caching and validated responses.
  - Added configuration options for price-history providers and upstream services.
- **Bug Fixes**
  - Improved handling of incomplete market-data responses and valid zero supply values.
  - Added consistent errors for failed upstream requests and invalid inputs.
- **Tests**
  - Added unit and integration coverage for price history, supply, fallback, validation, and caching.


kernelwhisperer:
@coderabbitai review

apps/api/src/app/routes/__chainId/tokens/__tokenAddress/priceHistory/codex.provider.ts
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  return bars.t.flatMap((timestamp, index) => {
+    const open = bars.o[index]
+    const high = bars.h[index]
+    const low = bars.l[index]
+    const close = bars.c[index]
+    const rawVolume = bars.volume?.[index]
+
+    if (open === null || high === null || low === null || close === null) {
```
azebuado, line 114: What happens if any of those is undefined? Can that happen?

kernelwhisperer: The candle is skipped because it's malformed, this ensures we don't throw at runtime, but it's defensive, it shouldn't happen.

The pattern that is commonly used within our code is to validate the responses coming from 3rd parties.

azebuado: The `null` check part I'm fine with, I was wondering about the list not having the given index resulting in those values being `undefined`.

apps/api/src/app/routes/__chainId/tokens/__tokenAddress/priceHistory/index.ts
```diff
@@ -0,0 +1,92 @@
...
+    providers.push(new UpstreamPriceHistoryProvider(fastify.config.PRICE_HISTORY_UPSTREAM))
+  } else {
+    fastify.log.warn('PRICE_HISTORY_UPSTREAM is not set. Upstream price history provider is disabled.')
+  }
+
+  if (fastify.config.CODEX_API_KEY) {
+    providers.push(new CodexPriceHistoryProvider(fastify.config.CODEX_API_KEY))
+  } else {
+    fastify.log.warn('CODEX_API_KEY is not set. Codex price history fallback is disabled.')
+  }
```
azebuado, line 39: A nitpick at this point, as I don't want to block on this potential refactor opportunity. What if we have a third provider? Would it make sense to have a registry similar to the USD price repository?

apps/api/src/app/routes/__chainId/tokens/__tokenAddress/priceHistory/codex.provider.ts
kernelwhisperer: We have 2 layers of validation:

1. hard throw if we get an invalid response 
image
2. if we get an incomplete candle (as per Codex convention this can happen due to different reasons) we skip that candle, but we don't throw


So if we get `undefined` bff returns 500, if we get `null` we filter those candles out.

apps/api/src/app/plugins/env.ts
azebuado, line 1: <details>
<summary>⚠️ <strong>AI Review</strong> (Claude Opus 5, worked 7m): one bad upstream datapoint discards the whole series; CORS handling differs between the two new endpoints</summary>

**Finding:** [BLOCKING] A single non-positive price from upstream throws away the entire price history response

This one is important:

- Location: `priceHistory/priceHistory.types.ts:51-63` (`normalizePriceHistoryBars` / `isValidBar`), reached from `priceHistory.service.ts:63`
- `upstreamPriceHistoryPayloadSchema` accepts `openUsd`/`highUsd`/`lowUsd`/`closeUsd` as plain `{ type: 'number' }` with no lower bound (`priceHistory.schemas.ts:129-132`), but `isValidBar` requires `price > 0` (`priceHistory.types.ts:71`). `normalizePriceHistoryBars` **throws** on the first bar that fails, so one zero-priced candle discards every other bar in the batch.
- The service catches that throw as "provider failed" and moves on, which means:
  - Codex not configured (the `.env.example` default — `CODEX_API_KEY` is commented out, and `index.ts:38` warns "Codex price history fallback is disabled"): the client gets **502 and no chart at all**.
  - Codex configured: the whole series is silently swapped to the fallback provider, so the chart jumps to different data because of one bad datapoint.
- I confirmed both by injecting the payload into the route (`app.inject`) at PR head:
  - upstream-only, candles `[{1,3,0.5,2.5}, {0,0,0,0}, {2,4,1,3}]` → `502 {"message":"Price history providers failed"}`
  - with Codex enabled, same upstream payload → `200 {"providerId":2, …}`, both valid upstream bars dropped
- This is also inconsistent with the rest of the pipeline: `codex.provider.ts:114` and `upstream.provider.ts:72` both deliberately *skip* unusable entries. Only the shared normalizer is all-or-nothing.

**Suggested fix**
- Make `normalizePriceHistoryBars` skip invalid bars instead of throwing, matching the providers' existing policy, and log the drop count so bad upstream data stays visible.
- If all-or-nothing is intentional, say so in the PR body and tighten `upstreamPriceHistoryPayloadSchema` to `exclusiveMinimum: 0` so the payload is rejected at the schema boundary rather than deep inside the normalizer.
- Test: one valid + one zero-priced upstream candle, asserting the surviving bars rather than a 502.

**Finding:** [QUESTION] `priceHistory` reflects the request origin, `supply` does not — why the difference?

- Location: `priceHistory/index.ts:43-50` vs `supply/index.ts` (no equivalent hook)
- `priceHistory` adds an `onSend` hook that echoes any `Origin` back as `Access-Control-Allow-Origin`, and `index.spec.ts:76` asserts it. The `supply` route added in the same PR has no CORS handling at all.
- The app-wide policy is `origin: false` with `preflight: false`, allowing only localhost (`plugins/cors.ts:5-21`). The consuming frontend fetches both endpoints from the browser against `BFF_BASE_URL` (`fetchPriceChartData.ts`, `fetchTokenSupply.ts` in cowprotocol/cowswap#8009), so exactly one of these is wrong:
  - If the reflection is what makes the endpoint reachable cross-origin, `/supply` is broken in the browser and the chart loses market cap.
  - If it is not needed — which is what I'd expect, since every other BFF route the frontend already calls has no such hook — then this is dead code that quietly widens one route's CORS past the app-wide policy, with nothing in the PR body noting it.
- Also worth noting: `proxies/socket/index.ts:20-33` is the one existing route doing per-route CORS, and it pairs the reflection with an explicit `OPTIONS` handler precisely because global `preflight` is off. `priceHistory` has the reflection without the handler, so anything that ever triggers a preflight here fails.

**Suggested fix**
- Either drop the hook (if the edge already handles CORS) or apply the same treatment to `/supply` plus an `OPTIONS` handler, and note the deviation from `origin: false` in the PR description.

**Finding:** [NON-BLOCKING] Price history *is* cached server-side, contrary to the PR description

- Location: `priceHistory/index.ts:83` (`max-age=30`) combined with `plugins/bffCache.ts:55-80`
- The PR body says "Price history has no server-side response cache. It only returns a 30-second HTTP cache header." `bffCache` is `fp`-wrapped and autoloaded from `plugins/` (`bffCache.ts:126-129`), so its hooks are global: `getTtlFromResponse` (`bffCache.ts:118-122`) reads `max-age` off the reply and any 2xx GET carrying one is written to the shared cache repository under `requests:<url>` (`getKey`, `bffCache.ts:114`).
- So every price history response lands in Redis for 30s. Two consequences worth a decision:
  - The documented design intent ("no server-side cache") is not what ships.
  - The key is the full URL including query string, so each distinct `from`/`to`/`interval`/`countback` tuple is its own entry — if the client's `to` tracks wall-clock rather than bar boundaries, that is one write-only Redis key per request.
- The new route specs register the plugin into a bare Fastify instance, so they can't surface this; it only appears in the assembled app.

**Suggested fix**
- Decide which behavior you want: keep the cache and correct the PR description, or skip the response cache for this route and document how.

<details>

Automated checks run at PR head `384e9e7`:
- `jest --config apps/api/jest.config.ts --testPathPattern '(priceHistory|supply)'` → 5 suites, 38 tests, green across repeated runs.
- `jest --listTests` confirms `*.int.ts` is not picked up by the nx preset's `testMatch`, so the live provider tests stay out of `yarn test`. Good.
- Targeted `app.inject` probes for the zero-price case above.

Separate from existing review comments, which already cover:
- **Countback window** (`upstream.provider.ts:95`, bot thread): still unresolved on head. My read differs from the suggestion in that thread — swapping `Math.min` for `Math.max` would *shrink* the window and return fewer bars than `countback` asked for, which is the wrong direction for TradingView-style countback semantics. The real asymmetry is that the upstream provider widens the window but never trims the result to `countback`, while Codex passes `countback` through and returns exactly that many, so the two providers answer the same request with different bar counts. `upstream.provider.spec.ts:42` only asserts length 1, so neither behavior is pinned. Worth deciding explicitly rather than applying the suggested flip.
- **Falsey supply entries** (`supply/index.ts:55`): fixed on head — the guard is `tokenSupply === undefined`, so `null`/`0`/`false` now reach schema validation instead of 404.
- **Codex array indexing** (`codex.provider.ts:114`): answered separately on that thread.

Not raised as findings: the per-provider 10s deadline in `priceHistory.service.ts:60` re-implements the existing `libs/repositories/src/utils/fetchWithTimeout.ts` (same 10s default), and the `supply` route's three bare `throw new Error()` plus `catch {}` mean upstream 5xx, timeouts, and schema drift all surface as an unlogged 502.

</details>

<details>

```text
Verify each finding against current code. Fix only if still valid, keep the change minimal, and validate with targeted tests.

Context:
1. apps/api/src/app/routes/__chainId/tokens/__tokenAddress/priceHistory/priceHistory.types.ts:51-63
   - normalizePriceHistoryBars throws on the first bar failing isValidBar, discarding the whole batch.
   - upstreamPriceHistoryPayloadSchema allows openUsd/highUsd/lowUsd/closeUsd as unbounded numbers, while
     isValidBar requires price > 0, so one zero-priced upstream candle yields 502 (Codex disabled) or a
     silent full-series switch to Codex.
   - Expected: skip invalid bars like codex.provider.ts:114 and upstream.provider.ts:72 already do, and log
     the drop count. Add a spec with one valid plus one zero-priced upstream candle asserting surviving bars.

2. priceHistory/index.ts:43-50 vs supply/index.ts
   - priceHistory reflects Origin into Access-Control-Allow-Origin; supply has no CORS handling. App-wide
     policy is origin:false, preflight:false (plugins/cors.ts).
   - Expected: make the two endpoints consistent — remove the hook, or apply it to supply together with an
     OPTIONS handler as proxies/socket/index.ts does.

3. priceHistory/index.ts:83 with plugins/bffCache.ts:55-122
   - bffCache is fp-wrapped and global; it caches any 2xx GET carrying max-age, so price history is cached
     server-side for 30s despite the PR body saying it is not.
   - Expected: either correct the PR description or opt this route out of the response cache.
```

</details>

*Generated using the `pr-review` skill from the [CoW Protocol skills repo](https://github.com/cowprotocol/skills).*

</details>

azebuado (approved):
Ok from me with a nitpick, AI review has some questions/comments

